PR #3, opened 2019-07-29
Reduce duplication of emitted elements

tsenart:
This commit reduces duplication of emitted elements which reduces the
size of the produced index as well as the exporting run-time by a large margin.

Better reviewed without whitespace changes.

On our sourcegraph/sourcegraph repo:

- Size: 1.7G -> **88M**
- Time: 27.5s -> **6.2s**

For the upload routine, we should additionally Gzip compress the index file,
which one the above index, results in a further reduction to **11M**.

### DIFF
--- a/export/exporter.go
+++ b/export/exporter.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"go/ast"
 	"go/token"
 	"go/types"
 	"io"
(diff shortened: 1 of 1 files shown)


unknwon:
Can’t wait to see how it goes for k8s repo with over 2.7M lines of Go!

unknwon:
Finished k8s repo in 1m26s with 816M w/o compression, w/o embedding file content.

PR #7, opened 2019-07-31
Remove useless duplicated hover result

### LINKED ISSUE #5: Lots of duplication
If you search for `type Router struct` in the output LSIF for gorilla/mux, you can see that the same hover result is duplicated 33 times, all referenced from the same result set.

Here's a visualisation:

image

The center is a single `resultSet` vertix, with many `textDocument/hover` edges that each point to duplicated hover result vertixes with the same content.

efritz:
These two lines created a hover result and linked it to the definition's result set id. These edges already exist from exportDef (maybe it was originally written to link the use result id to the hover result id, but that's not the behavior).

This reduces file size slightly (1.5M -> 1.1M for gorilla/mux) but also reduces edges that go nowhere and edges that we'll have to query via json/sqlite/dgraph at runtime.

This closes #5.

### DIFF
--- a/export/exporter.go
+++ b/export/exporter.go
@@ -397,16 +397,6 @@ func (e *exporter) exportUses(p *packages.Package, fi *fileInfo, filename string
 			return fmt.Errorf(`emit "item": %v`, err)
 		}
 
-		hoverResultID, err := e.emitHoverResult(def.contents)
-		if err != nil {
-			return fmt.Errorf(`emit "hoverResult": %v`, err)
-		}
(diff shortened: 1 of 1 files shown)


unknwon:
>(maybe it was originally written to link the use result id to the hover result id, but that's not the behavior

Originally, we found that top level decls without any uses inside the package won't show the hover result. Then realize we need a hover vertex back to definition itself, which seems make these hoverResult no longer necessary.

Good catch!

PR #8, opened 2019-07-31
Convert from string to int identifiers.

efritz:
The lsif-tsc utility uses either strings or uuid strings, not int-like strings. This change will make our implementation mirror that a bit more closely (with some small space savings during transmission as well).

### DIFF
--- a/export/exporter.go
+++ b/export/exporter.go
@@ -12,7 +12,6 @@ import (
 	"io/ioutil"
 	"log"
 	"path/filepath"
-	"strconv"
 
 	doc "github.com/slimsag/godocmd"
 	"github.com/sourcegraph/lsif-go/protocol"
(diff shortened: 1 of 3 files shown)


unknwon:
The original idea for using `string` type ID is to have the possibility for easy conversion to other algorithms, which can be used to compute the vertex/edge ID, e.g. SHA1, UUID.

My assumption is to use globally unique identifier "could" help future work on [package linker](https://github.com/microsoft/language-server-protocol/blob/master/indexFormat/implementation.md#package-linker), which "may" make thing easier when bridge references and definitions across different LSIF dumps. Since the ID algorithm used in current version is simply start from 1 for every dump, this algorithm will no longer in package linker. 

And, if the ID must be a string in package linker, whether or not we need to change ID generate algorithm, package `protocol` can be reused.

However, I haven't had time investigate on how package linker is supposed to work, and how "moniker" type vertex/edge should be emitted, my assumption can be totally wrong.

unknwon:
cc @chrismwendt

unknwon:
Quote from @tsenart:

>Using a hash of certain information contained in each element instead of integer IDs could help us detect duplication with the ids alone.
With that, we could pipe the index file to  `| jq -c '.id' | uniq -c | sort -nr` to detect unwanted duplication.

efritz:
That makes sense. Savings would be minimal anyway under gzip.

PR #11, opened 2019-07-31
Emit events

efritz:
These events are required for lsif-sqlite to properly be able to convert the dumps into a database.

### DIFF
--- a/protocol/protocol.go
+++ b/protocol/protocol.go
@@ -135,7 +135,7 @@ func NewEvent(id, kind, scope string, data string) *Event {
 		},
 		Kind:  kind,
 		Scope: scope,
-		Data:  data, // TODO - will need to be int soon
+		Data:  data,
 	}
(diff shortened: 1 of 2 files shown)


protocol/protocol.go
```diff
@@ -110,6 +110,35 @@ func NewMetaData(id, root string, info ToolInfo) *MetaData {
...
+	Kind string `json:"kind"`
+	// The type of element this event describes (project or document).
+	Scope string `json:"scope"`
+	// The identifier of the data beginning or ending.
+	Data string `json:"data"`
+}
+
+// NewEvent returns a new Event object with the given ID, kind, scope,
+// and data information.
+func NewEvent(id, kind, scope string, data string) *Event {
```
unknwon:
```suggestion
func NewEvent(id, kind, scope, data string) *Event {
```

PR #12, opened 2019-07-31
Remove duplicate document end $event vertices.

efritz:
This fixes a problem where the size of e.files grows and the number of times a document is output is the number of files processed after it.

### DIFF
--- a/export/exporter.go
+++ b/export/exporter.go
@@ -119,6 +119,19 @@ func (e *exporter) export(info protocol.ToolInfo) error {
 		}
 	}
 
+	// Close all documents
+	// TODO(efritz) - see if we can rearrange the outputs so that
+	// all of the output for a document is contained in one segment
+	// that does not interfere with emission of other document
+	// properties.
+
+	for _, info := range e.files {
+		_, err = e.emitEndEvent("document", info.docID)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


export/exporter.go
```diff
@@ -119,6 +119,19 @@ func (e *exporter) export(info protocol.ToolInfo) error {
 		}
 	}
 
+	// Close all documents
+	// TODO(efritz) - see if we can rearrange the outputs so that
+	// all of the output for a document is contained in one segment
+	// that does not interfere with emission of other document
+	// properties.
+
+	for _, info := range e.files {
```
unknwon: I think this loop can be merged to https://github.com/sourcegraph/lsif-go/pull/12/files#diff-14e51e63d8b1ecbb10fbc9c91d32bef6R89-R120, they both iterate over `e.files`. (just found out this locally while working on something else)

efritz: Fixed

efritz: Sorry, reverted the fix. That doesn't work as we need to emit all of the document end events after everything about that document is done. This will spread them out too far and makes it behave strangely with lsif-sqlite.

unknwon: I see... maybe add another note saying something like "previous loop could contain cross-file references and it is only accurate to emit end event in a separate loop afterwards"?

Issue #10, opened 2019-07-31
No LSIF info for uses found before defs


unknwon:
### Description

Missing `textDocument/references` vertex for uses (aka references) in a file found before the file contains that definition.

This regression is introduced in https://github.com/sourcegraph/lsif-go/commit/ba053fa620c9fe1d77c40e1ff0649a256ee2f90c.

e.g.
- [`Route`](https://github.com/gorilla/mux/blob/e67b3c02c7195c052acff13261f0c9fd1ba53011/mux.go#L25) is used in `mux.go` but its definition is only found in the file `route.go` which is processed after `mux.go`

```
Package: mux
	...
	File: .../mux.go
	...
	File: .../route.go
...
```

### Observation

<details><summary>Actual</summary>

![2019-07-31 13 24 26](https://user-images.githubusercontent.com/2946214/62245461-89ceab80-b396-11e9-93b0-e70c9f2387f6.gif)
</details>

<details><summary>Expected</summary>

![2019-07-31 13 25 18](https://user-images.githubusercontent.com/2946214/62245531-abc82e00-b396-11e9-9516-ce1e75941dab.gif)
</details>

unknwon:
Closed by #10.

PR #13, opened 2019-07-31
export/exporter: emit refs after all files are processed in a package (#10)

unknwon:
Emit refs on each file basis will lose connections to defs defined in files that have not yet processed.

Finished k8s/k8s repo in 1m32s with 1.1G (was 816M after #3) w/o compression, w/o embedding file content.

Context: #10

### DIFF
--- a/export/exporter.go
+++ b/export/exporter.go
@@ -162,9 +162,9 @@ func (e *exporter) exportPkg(p *packages.Package, proID string) (err error) {
 		}
 	}
 
-	// NOTE: since we currently only support package-level references, it is OK to run the loop
-	//       here. Once we want repository (or Go module)-level references, they can only be
-	//       handled after all files are processed for definitions.
+	// NOTE: since we currently only support package-level references, it is OK to export usages
+	// at the end of each package. When repository-level references are implemented, usages must
+	// be exported after all files are processed.
 	for _, f := range p.Syntax {
 		fpos := p.Fset.Position(f.Package)
 		if err := e.exportUses(p, e.files[fpos.Filename], fpos.Filename); err != nil {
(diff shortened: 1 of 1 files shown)


export/exporter.go
```diff
@@ -160,8 +160,14 @@ func (e *exporter) exportPkg(p *packages.Package, proID string) (err error) {
 		if err = e.exportDefs(p, f, fi, proID, fpos.Filename); err != nil {
 			return fmt.Errorf("error exporting definitions of %q: %v", p.PkgPath, err)
 		}
+	}
 
-		if err := e.exportUses(p, fi, fpos.Filename); err != nil {
+	// NOTE: since we currently only support package-level references, it is OK to run the loop
+	//       here. Once we want repository (or Go module)-level references, they can only be
+	//       handled after all files are processed for definitions.
```
chrismwendt: How about s/run the loop here/export usages at the end of each package/ and "When repository-level references are implemented, usages must be exported after all files are processed"? I'm not 100% sure if that's correct.

unknwon: I think your description is much better!

tsenart:
Also, how much is it compressed? For all intents and purposes, is there any reason for us to work with uncompressed LSIF indexes ever?

chrismwendt:
Test suite plans: https://docs.google.com/document/d/1cq9AA_kmDu2g1I4HI9J678yH0seKeW1ooecYmJqcpoo/edit?usp=sharing (you've seen this already)

Compression: compression for the newline-delimited JSON format simply hasn't been implemented yet. @efritz is investigating the performance implications (size and speed) of different backends like SQLite (see https://github.com/sourcegraph/sourcegraph/pull/5007).

unknwon:
@tsenart no worries! I have to admit that my mind wasn't so clear when I did the code review. 

>Also, how much is it compressed? For all intents and purposes, is there any reason for us to work with uncompressed LSIF indexes ever?

I think mainly (same reason LSIF spec team choose JSONLines) because it's early stage and easy for developers to start working on it with plain text format.

Indexer can also do GZIP directly (i.e. `-gzip`), but if `gzip` command works so fast, it's not worth it.

PR #16, opened 2019-08-05
Duplicate edges

efritz:
This removes duplicated contains edges which messes up the SQLite conversion utility.

### DIFF
--- a/export/exporter.go
+++ b/export/exporter.go
@@ -476,17 +476,26 @@ func (e *exporter) exportUses(p *packages.Package, fi *fileInfo, filename string
 			return fmt.Errorf(`emit "next": %v`, err)
 		}
 
-		defResultID, err := e.emitDefinitionResult()
-		if err != nil {
-			return fmt.Errorf(`emit "definitionResult": %v`, err)
+		// If this is the first use for this definition, we need to create
+		// some extra vertices. Caching this on the definition lets us share
+		// the vertices between uses. We do this lazily so that we don't have
+		// an unreachable set of vertices.
+
+		if def.defResultID == "" {
+			defResultID, err := e.emitDefinitionResult()
+			if err != nil {
+				return fmt.Errorf(`emit "definitionResult": %v`, err)
+			}
+
+			_, err = e.emitTextDocumentDefinition(def.resultSetID, defResultID)
+			if err != nil {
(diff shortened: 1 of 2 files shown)


export/exporter.go
```diff
@@ -150,6 +150,20 @@ func (e *exporter) export(info protocol.ToolInfo) error {
 	return nil
 }
 
+func deduplicate(values []string) []string {
```
unknwon: The use of **append** ([on line 123](https://github.com/sourcegraph/lsif-go/pull/16/files#diff-14e51e63d8b1ecbb10fbc9c91d32bef6R123)) seems no longer necessary with use of this new **deduplicate** function, since we could accept two `[]string` then merge them into one.

export/exporter.go
```diff
@@ -120,7 +120,7 @@ func (e *exporter) export(info protocol.ToolInfo) error {
 		}
 
 		if len(f.defRangeIDs) > 0 || len(f.useRangeIDs) > 0 {
-			_, err = e.emitContains(f.docID, append(f.defRangeIDs, f.useRangeIDs...))
+			_, err = e.emitContains(f.docID, deduplicate(append(f.defRangeIDs, f.useRangeIDs...)))
```
unknwon: It would be better if we have a comment about why duplications would happen here.

export/types.go
```diff
@@ -24,6 +24,8 @@ type defInfo struct {
 	resultSetID string
 	// The contents will be used as the hover information.
 	contents []protocol.MarkedString
+	// The lazily initialized definition result ID
```
unknwon:
```suggestion
	// The lazily initialized definition result ID upon first use found.
```

unknwon: Aside: my English wording can be wrong, use your best judge!

export/exporter.go
efritz: I use append after truncating so we can reuse the same allocated space

export/exporter.go
efritz: Will do

export/exporter.go
unknwon: Use `strs := make([]string, s1, len(s1)+len(s2))` inside this function should be effectively same as has an **append** beforehand.

efritz:
Added a description.

PR #23, opened 2019-08-20
Add moniker support

efritz:
This PR 

- Adds moniker output to the lsif-go binary
- Adds a new utility to add read go.mod and go.sum to and rewrite monikers for external use
- Adds support for stdout output for lsif-go (this allows us to stream lsif-go output directly to lsif-gomod)
- Refactors the binary arguments to act like lsif-tsc (now that we have two cmds, this was a good opportunity)


unknwon:
Haven't looked in to the actual implementation, just want to say the package structure was designed to use subcommands within a single binary to avoid producing multiple binaries for easy distribution. That's why the equivalent command to `lsif-tsc` is `lsif-go export` not `lsif-go`, and `lsif-go link` was the design/idea for the package linker (sorry didn't put this into the doc somewhere). With that said, the subcommand approach is not necessary good. cc @chrismwendt

felixfbecker:
I think either the monikers should always be added, or it should be a separate tool (not just a subcommand). The thought behind separate tools is that you may use different package managers, and you have one tool per package manager. I don't know if that applied to Go, or if everyone is already using Go modules?

unknwon:
>The thought behind separate tools is that you may use different package managers, and you have one tool per package manager. I don't know if that applied to Go, or if everyone is already using Go modules?

The current source code parser `go/packages` behaves the exact same way as `go` command which takes care of build environment, it is aware of where to look for dependencies, GOPATH, vendor and/or Go modules (i.e. `GOPATH/pkg/mod`). 

If the user wants to be absolutely explicit about which mechanism for looking up dependencies, we could also have something like `lsif-go link -driver=gomod` or `lsif-go link -driver=vendor`.

felixfbecker:
Aren't there other package managers than gomod (and vendoring)? How do you figure out the version for a vendored module?

unknwon:
>Aren't there other package managers than gomod (and vendoring)? How do you figure out the version for a vendored module?

No matter which vendor/pm tool a project/package uses, ultimately it has to be able to compiled by "go" command. And if you're talking about the dependency version info extraction (e.g. `Gopkg.lock`, `vendor.json`), then the `lsif-go link -driver=xx` would still work.

efritz:
> I think either the monikers should always be added, or it should be a separate tool.

The separate tool is the recommendation by Microsoft, which I think we should follow.

unknwon:
Just to be clear, I'm fine with either separate binaries or subcommands.

cmd/lsif-go/main.go
```diff
@@ -0,0 +1,100 @@
+// The program lsif-go is an LSIF exporter for Go.
```
unknwon: Should we change this "exporter" to "indexer"?

cmd/lsif-gomod/main.go
```diff
@@ -0,0 +1,79 @@
+package main
```
unknwon: Do we need a line of comment for description of what lsif-gomod is/does?

cmd/lsif-gomod/main.go
```diff
@@ -0,0 +1,79 @@
...
+func realMain() error {
+	var (
+		projectRoot string
+		inFile      string
+		stdin       bool
+		outFile     string
+		stdout      bool
+	)
+
+	app := kingpin.New("lsif-go", "lsif-gomod decorates lsif-go output with gomod monikers.").Version(versionString)
```
unknwon:
```suggestion
	app := kingpin.New("lsif-gomod", "lsif-gomod decorates lsif-go output with gomod monikers.").Version(versionString)
```

export/exporter.go
```diff
@@ -591,3 +590,13 @@ func (e *exporter) emitItemOfReferences(outV string, inVs []string, docID string
 	id := e.nextID()
 	return id, e.emit(protocol.NewItemOfReferences(id, outV, inVs, docID))
 }
+
+func (e *exporter) emitMoniker(kind, sourceID, identifier string) error {
+	monikerID := e.nextID()
+	err := e.emit(protocol.NewMoniker(monikerID, kind, "go", identifier))
```
unknwon:
```suggestion
	err := e.emit(protocol.NewMoniker(monikerID, kind, protocol.LanguageID, identifier))
```

~~Or if it's always "go", we could reduce one parameter and set this value inside `protocol.NewMoniker`.~~

gomod/moniker.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+}
+
+func (d *decorator) decorate(line string) error {
+	moniker := &moniker{}
+	if err := json.Unmarshal([]byte(line), &moniker); err != nil {
+		return fmt.Errorf("unmarshal line: %v", err)
+	}
+
+	if moniker.Type == "vertex" && moniker.Label == "moniker" {
```
unknwon:
```suggestion
	if moniker.Type == protocol.ElementVertex && moniker.Label == protocol.VertexMoniker {
```

protocol/protocol.go
```diff
@@ -357,6 +357,60 @@ func NewHoverResult(id string, contents []MarkedString) *HoverResult {
...
+	Vertex
+	// The name of the package.
+	Name string `json:"name"`
+	// The package manager.
+	Manager string `json:"manager"`
+	// The version of the package.
+	Version string `json:"version"`
+}
+
+// NewPackageInformation returns a new PackageInformation with the given ID< name, manager, and version.
```
unknwon:
```suggestion
// NewPackageInformation returns a new PackageInformation with the given ID, name, manager, and version.
```

cmd/lsif-gomod/main.go
efritz: Done.

cmd/lsif-go/main.go
efritz: Will change all of these in another PR.

PR #24, opened 2019-09-17
Fix bad document field value on item edges.

efritz:
The `document` property on `item` edges refers to the containing document of the `inVs` values, not the `outV` value. @chrismwendt and I found this out the hard way playing with the C++ indexer.

### DIFF
--- a/export/exporter.go
+++ b/export/exporter.go
@@ -121,7 +121,7 @@ func (e *exporter) export(info protocol.ToolInfo) (*Stats, error) {
 				}
 			}
 
-			for docID, rangeIDs := range e.refs[rangeID].defRangeIDs {
+			for docID, rangeIDs := range e.refs[rangeID].refRangeIDs {
 				_, err = e.emitItemOfReferences(refResultID, rangeIDs, docID)
 				if err != nil {
 					return nil, fmt.Errorf(`emit "item": %v`, err)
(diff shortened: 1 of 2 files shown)


export/exporter.go
```diff
@@ -114,13 +114,15 @@ func (e *exporter) export(info protocol.ToolInfo) (*Stats, error) {
...
+			for docID, rangeIDs := range e.refs[rangeID].defRangeIDs {
+				_, err = e.emitItemOfDefinitions(refResultID, rangeIDs, docID)
+				if err != nil {
+					return nil, fmt.Errorf(`emit "item": %v`, err)
+				}
 			}
 
-			if len(e.refs[rangeID].refRangeIDs) > 0 {
-				_, err = e.emitItemOfReferences(refResultID, e.refs[rangeID].refRangeIDs, f.docID)
+			for docID, rangeIDs := range e.refs[rangeID].defRangeIDs {
```
chrismwendt: Should this still be `refRangeIDs`?

export/types.go
```diff
@@ -14,6 +14,9 @@ type fileInfo struct {
 
 // defInfo contains LSIF information of a definition.
 type defInfo struct {
+	// The identifier of the containing document. This is necessary
+	// to track when emitting item edges sthat
```
unknwon: What is "sthat"?

efritz: Sentence wasn't finished :p

export/exporter.go
efritz: Fixed.

PR #28, opened 2019-11-20
Fix monikers

efritz:
We were not emitting any data about uses that were not defined inside the package. This change:

- ensures monikers are emitted for such uses
- standardizes moniker format with identifiers (`package:ident`)
- correlates monikers correctly with the package from go.sum (previously required an exact match, not a prefix match)
- does not hardcode a module version of 0.0.0 for defs

### DIFF
--- a/cmd/lsif-go/main.go
+++ b/cmd/lsif-go/main.go
@@ -50,12 +50,12 @@ func realMain() error {
 		return fmt.Errorf("either an output file using --out or --stdout must be specified")
 	}
 
-	if debug {
-		log.SetLevel(log.Debug)
-	}
-
 	if verbose {
 		log.SetLevel(log.Info)
+	}
+
+	if debug {
+		log.SetLevel(log.Debug)
 	}
 
 	if stdout && (verbose || debug) {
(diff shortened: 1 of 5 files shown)


cmd/lsif-gomod/main.go
```diff
@@ -50,6 +52,12 @@ func realMain() error {
 		return fmt.Errorf("either an output file using --out or --stdout must be specified")
 	}
 
+	if moduleVersion == "" {
+		// For future reference, see
+		// https://github.com/golang/vgo/blob/9d567625acf4c5e156b9890bf6feb16eb9fa5c51/vendor/cmd/go/internal/modfetch/coderepo.go#L88
+		return fmt.Errorf("module version cannot be inferred")
```
chrismwendt:
```suggestion
		return fmt.Errorf("module version cannot be inferred so set it explicitly with -moduleVersion=...")
```

cmd/lsif-go/main.go
```diff
@@ -50,14 +50,14 @@ func realMain() error {
...
-		log.SetLevel(log.Debug)
-	}
-
 	if verbose {
 		log.SetLevel(log.Info)
 	}
 
+	if debug {
+		log.SetLevel(log.Debug)
+	}
```
unknwon: Quick question: was this change intentional? (i.e. show logs in debug level if both `verbose` and `debug` are `true`)

efritz: `-verbose` will set Info, `-debug` will set Debug and will override the -verbose flag. The order they were meant if you do `-v -debug`, then you only get info output. Now specifying verbose and debug will take the most specific level.

PR #29, opened 2019-11-21
Fix external refs

efritz:
lsif-server makes the assumption (which I now believe to be correct after reflecting a bit more on it) that all references will be the inV of a reference item edge. *Whether or not* the range/resultSet also has a reachable definition is an orthogonal concern.

See the following image:

reference results

I believe this implies that if a def is external to the project, that a moniker can be attached to the range or resultSet, but the rest of the graph should remain the same (both the `textDocument/references` as well as the `item[ref]` edge should exist).

This PR creates those edges. A future PR will try to combine some of these reference results in order to save space.

### DIFF
--- a/index/indexer.go
+++ b/index/indexer.go
@@ -465,9 +465,10 @@ func (e *indexer) indexUses(p *packages.Package, fi *fileInfo, filename string) 
 				return fmt.Errorf(`emit moniker": %v`, err)
 			}
 
-			// UHHHHHHHHHHHHH
-			// TODO - explain
-			// UHHHHHHHHHHHHH
+			// Emit a reference result edge and create a small set of edges that link
+			// the reference result to the range (and vice versa). This is necessary to
+			// mark this range as a reference to _something_, even though the definition
+			// does not exist in this source code.
 
 			refResultID, err := e.emitReferenceResult()
 			if err != nil {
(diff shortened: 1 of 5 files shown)


index/indexer.go
```diff
@@ -464,6 +465,26 @@ func (e *indexer) indexUses(p *packages.Package, fi *fileInfo, filename string)
...
+			// the reference result to the range (and vice versa). This is necessary to
+			// mark this range as a reference to _something_, even though the definition
+			// does not exist in this source code.
+
+			refResultID, err := e.emitReferenceResult()
+			if err != nil {
+				return fmt.Errorf(`emit "referenceResult": %v`, err)
+			}
+
+			_, err = e.emitTextDocumentReferences(rangeID, refResultID)
```
chrismwendt: Does the spec say

`range->referenceResult`

is a valid relationship? I thought it had to go through a resultSet:

`range->resultSet->referenceResult`

Are you thinking of monikers? I recall us chatting about that yesterday. Wait a second, is

`range->moniker`

even a valid relationship? Huh, the spec says in prose it's valid:

> ...LSIF introduces optional monikers which can be linked to ranges using a corresponding edge.
> ...with a moniker (e.g. a handle in string format) that is bound to the corresponding range declaration.

But all of the examples use a `resultSet` between the range and moniker. 😕

PR #30, opened 2019-11-21
Module versions

efritz:
- Infer the module version
- Cut out dates from pseudo-versions as they are client-dependent and will certainly mismatch if used as monikers
- Use go list commands rather than reading go.mod,go.sum files directly
- rename package module where applicable


gomod/module.go
```diff
@@ -0,0 +1,100 @@
...
+		mostRecentTag = "v0.0.0"
+	}
+
+	//f Step 3: Determine the current commit and suffix it with the most
+	// recent tag.
+	commit, err := run(projectRoot, "git", "rev-parse", "HEAD")
+	if err != nil {
+		return "", fmt.Errorf("failed to get current commit: %v", err)
+	}
+	return fmt.Sprintf("%s-%s", mostRecentTag, commit[:12]), nil
```
chrismwendt: Is there a separate place for the discussion about how to handle cross-repo defs/refs when no LSIF data exists for the exact package@version? A few points we discussed today:

- Don't want lsif-server to include language-specific logic for parsing version strings
- Might be able to encode the version as a lexicographically-sortable string to find the nearest package@version (perhaps encoding could be done in a separate LSIF filter, like lsif-npm)
- V0 could be: fall back to visible_at_tip if no dump defines the exact package@version

gomod/module.go
```diff
@@ -0,0 +1,100 @@
...
+	"bytes"
+	"fmt"
+	"os/exec"
+	"regexp"
+	"strings"
+)
+
+// InferModuleVersion determines the go module version that is supplied by the
+// code in the given project root.
+func InferModuleVersion(projectRoot string) (string, error) {
```
unknwon: Just a note, not 100% sure if this function aware/handles this: Go modules does not allow v2.x.x tag for module path without containing a `/v2`, similar for `v3`, `v4` etc.

Eg. `module github.com/sourcegraph/sourcegraph` is only allowed to use tag v0 and up to v1. If try to use `v2` the Go modules mode won't allow user to compile. `module github.com/sourcegraph/sourcegraph/v2` is the module path Go modules toolchain expects to see when using tags of v2.x.x.

Even if we publish `module github.com/sourcegraph/sourcegraph` starting at v2 (no v1 or v0 whatsoever), it won't work AFAIK.

PS: Modules has import path `gopkg.in/xxx` are special cases that `xxx.v2` could work (not `xxx/v2`).

unknwon: If we haven't handled this, I don't think we have to take care of it now, assumes users would have fixed the compile issue before pushing to the code host.

gomod/module.go
efritz: Discussion in [RFC 82](https://docs.google.com/document/d/1hLyqyvhBUax4zIkRLV-sAreGSx49EAcyPrweIQtjMUQ).

gomod/module.go
efritz: This is good info. Could you make an issue so that we remember to handle this at some point?

PR #32, opened 2019-11-27
Print progress

efritz:


index/indexer.go
```diff
@@ -22,7 +23,7 @@ import (
 // Index generates an LSIF dump for a workspace by traversing through source files
 // and storing LSP responses to output source that implements io.Writer. It is
 // caller's responsibility to close the output source if applicable.
-func Index(workspace string, excludeContent bool, w io.Writer, moduleName, moduleVersion string, dependencies map[string]string, toolInfo protocol.ToolInfo) (*Stats, error) {
+func Index(workspace string, excludeContent bool, w io.Writer, moduleName, moduleVersion string, dependencies map[string]string, toolInfo protocol.ToolInfo, printProgressDots bool) (*Stats, error) {
```
unknwon: I feel the number of arguments here is too many, should we do something about it?

efritz: What? They're all necessary and they're basically direct results from cli input. I'm not sure it's worth it, it's only called once from main.

unknwon: > They're all necessary and they're basically direct results from cli input.

Technically every argument is necessary by design.

I'm thinking about something like a) `IndexOptions` or b) make a `type Indexer struct` and make this function be a method of the `Indexer`.

But this is not a blocker of this PR.

efritz: Updated anyway.

Issue #9, opened 2019-07-31
No j2d for defs with no uses


unknwon:
### Description

The `textDocument/definition` vertex (not 100% sure about the type) is not emitted for defs with no uses inside the same package.

Examples:
- https://github.com/gorilla/mux/blob/e67b3c02c7195c052acff13261f0c9fd1ba53011/mux.go#L24
- https://github.com/gorilla/mux/blob/e67b3c02c7195c052acff13261f0c9fd1ba53011/mux.go#L173

### Observation

<details><summary>Actual</summary>

![2019-07-31 13 02 05](https://user-images.githubusercontent.com/2946214/62244040-6b1ae580-b393-11e9-8d86-30516fb774c2.gif)
</details>

<details><summary>Expected</summary>

![2019-07-31 13 02 39](https://user-images.githubusercontent.com/2946214/62244065-7d951f00-b393-11e9-88b9-492014cf6532.gif)
</details>

PR #36, opened 2020-02-12
Always emit a definition result when the definition vertex is emitted.

efritz:
This should fix https://github.com/sourcegraph/lsif-go/issues/9.

I'm testing this by hand a little more rigorously tomorrow, so please don't merge if approved.


efritz:
@unknwon Do you know why it was like this originally? There's a few code layout things that look like there's a particular reason but I'm unsure what it is or if there is one for sure.

unknwon:
I'm not sure, maybe because of this? https://github.com/sourcegraph/lsif-go/commit/972f77bda0c9cbbc49a72e557727a9abb644d403

Issue #27, opened 2019-10-17
No LSIF data dumped for test files?


blanet:
LSIF dumps for test files should be generated and now they are missing.

Maybe Using `golang.org/x/tools/go/packages` is not the best way to dump LSIF considering generation speed and idempotent. But for now I have no idea how to improve this either. ;-)

efritz:
This looks like it was fixed in https://github.com/sourcegraph/lsif-go/commit/0e6f7dd01a88c649a0f5dc58bdead2a24a1dd2bc.

PR #37, opened 2020-02-12
Do not emit duplicate ranges

efritz:
It turns out that just following `package.Syntax` will have you visit the same document multiple times.

After this PR, the order will be:

- emit all document elements
- emit all def elements
- emit all ref elements
- emit all contains and item edges

We do this so that we can have a set of "whitelisted" documents while emitting document elements. This ensures that the output does not have documents from generated code (go test will create a slimmed down version of the code and re-index it).

We then maintain the set of filenames that we've emitted defs and uses for so we only visit them the first time in the sequence.

Lastly, we maintain a map of (filename, offset) pairs so that we do not emit a duplicate range, which happens mostly when a def and use shares an offset in a file. We can then just use the same range here.

After this change the output passes our validation util. As we do emit many fewer elements than before, I am going to re-benchmark what's on our LSIF documentation to see if there's a significant benefit here.


index/indexer.go
```diff
@@ -204,35 +235,71 @@ func (e *indexer) index(pkgs []*packages.Package) (*Stats, error) {
...
 
-func (e *indexer) indexPkg(p *packages.Package, proID string) (err error) {
-	log.Infoln("Package:", p.Name)
+func (e *indexer) indexPkgDocs(p *packages.Package, proID string) (err error) {
+	log.Infoln("Emitting documents for package", p.Name)
 	defer log.Infoln()
 
 	for _, f := range p.Syntax {
 		fpos := p.Fset.Position(f.Package)
+		if !strings.HasPrefix(fpos.Filename, e.projectRoot) {
```
unknwon: Not a blocker, but feels like a regression? https://github.com/sourcegraph/lsif-go/issues/27

unknwon: Should have commented on the next line:

>// Indexing test files means that we're also indexing the code generated by go test;

efritz: These files won't ever be served by the LSIF backend. I will verify by hand that test files still have support (that was originally enabled by setting `Test:true` when loading the package).

unknwon: Maybe just the comment itself confused me 😁

efritz: Still indexes test files.

PR #38, opened 2020-02-17
Move protocol writer into public package

efritz:
This will allow other utils like lsif-semanticdb to use the same helper methods to generate LSIF that aren't go index specific. This may be extracted into a separate library later, or just left here (the repo name is still pretty good for this).


protocol/writer.go
```diff
@@ -0,0 +1,156 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"strconv"
+)
+
+type Writer struct {
```
unknwon: Would be great to have docstring for this struct.

PR #41, opened 2020-02-26
Emit hover results for externally defined uses

efritz:
This gives us hover text for things we can't jump to such as `time.Millisecond`, `http.Request`, etc. This is a huge benefit, but still needs some work. Things that aren't defined at the top level still don't resolve, but I think it's possible with a little more elbow grease.


unknwon:
Thanks! This is already a huge step forward!

PR #51, opened 2020-03-27
Add a flag to specify an alternate module root.

creachadair:
Although it is common practice, not all Go repositories have the go.mod file at
the repo root. Add a --moduleRoot flag that allows the caller to specify a path
relative to the repository root, that should be used as the module root for the
purpose of indexing.

### DIFF
--- a/cmd/lsif-go/main.go
+++ b/cmd/lsif-go/main.go
@@ -105,9 +105,9 @@ func realMain() error {
 		Args:    os.Args[1:],
 	}
 
-	// TODO(@creachadair): Packages built with cgo may not include assembly
-	// files.  To index such a package you need to set CGO_ENABLED=0.  Consider
-	// maybe doing this explicitly, always.
+	// TODO(@creachadair): With cgo enabled, the indexer cannot handle packages
+	// that include assembly (.s) files.  To index such a package you need to
+	// set CGO_ENABLED=0.  Consider maybe doing this explicitly, always.
 	indexer := index.NewIndexer(
 		projectRoot,
 		repositoryRoot,
--- a/cmd/lsif-go/main.go
+++ b/cmd/lsif-go/main.go
@@ -17,7 +17,7 @@ import (
 	"github.com/sourcegraph/lsif-go/protocol"
 )
 
-const version = "0.4.1"
+const version = "0.5.0"
 const versionString = version + ", protocol version " + protocol.Version
 
 func init() {
--- a/cmd/lsif-go/main.go
+++ b/cmd/lsif-go/main.go
@@ -106,7 +106,7 @@ func realMain() error {
 	}
 
 	// TODO(@creachadair): With cgo enabled, the indexer cannot handle packages
-	// that include assembly (.s) files.  To index such a package you need to
+	// that include assembly (.s) files. To index such a package you need to
 	// set CGO_ENABLED=0. Consider maybe doing this explicitly, always.
 	indexer := index.NewIndexer(
 		projectRoot,
--- a/cmd/lsif-go/main.go
+++ b/cmd/lsif-go/main.go
@@ -81,6 +81,11 @@ func realMain() error {
 		return errors.New("module root is outside the repository")
 	}
 
+	// Ensure all the dependencies of the specified module are cached.
+	if err := gomod.Download(projectRoot); err != nil {
+		return fmt.Errof("fetching dependencies: %v", err)
+	}
+
 	moduleName, dependencies, err := gomod.ListModules(projectRoot)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


cmd/lsif-go/main.go
```diff
@@ -96,6 +105,9 @@ func realMain() error {
 		Args:    os.Args[1:],
 	}
 
+	// TODO(@creachadair): Packages built with cgo may not include assembly
+	// files.  To index such a package you need to set CGO_ENABLED=0.  Consider
+	// maybe doing this explicitly, always.
```
efritz: Oof you're a two spaces guy?

internal/gomod/module.go
```diff
@@ -60,7 +60,7 @@ func ListModules(projectRoot string) (string, map[string]string, error) {
 		return "", nil, nil
 	}
 
-	out, err := run(projectRoot, "go", "list", "-m", "all")
+	out, err := run(projectRoot, "go", "list", "-mod=readonly", "-m", "all")
```
efritz: Why this addition? Will this require that mod already be fetched?

cmd/lsif-go/main.go
```diff
@@ -47,6 +50,7 @@ func realMain() error {
 	app.Flag("out", "The output file.").Short('o').Default("dump.lsif").StringVar(&outFile)
 	app.Flag("moduleVersion", "Specifies the version of the module defined by this project.").PlaceHolder("version").StringVar(&moduleVersion)
 	app.Flag("repositoryRoot", "Specifies the path of the current repository (inferred automatically via git).").PlaceHolder("root").StringVar(&repositoryRoot)
+	app.Flag("moduleRoot", "Specifies the module root directory relative to the repository").Default(".").StringVar(&moduleRoot)
```
efritz: What is the difference between doing this and running the indexer in the correct root instead?

cmd/lsif-go/main.go
```diff
@@ -96,6 +105,9 @@ func realMain() error {
 		Args:    os.Args[1:],
 	}
 
+	// TODO(@creachadair): With cgo enabled, the indexer cannot handle packages
```
efritz: 😛

internal/gomod/module.go
creachadair: Without this, a package with vendored dependencies can't be listed. It should be harmless for any package without vendored dependencies. It does require that the dependencies be already loaded, but if that isn't the case then the indexer mutates the module cache which seems wrong.

We could add a `go mod download` step first, though.

cmd/lsif-go/main.go
creachadair: I'm not deeply partisan about it. It's just how I learned to type.

internal/gomod/module.go
efritz: To the tool or to the CI step?

cmd/lsif-go/main.go
creachadair: It doesn't work: If you run it from a subdirectory of the repository, the tool uses `git` to find the real root and looks for the module there. Similarly if you set `--repositoryRoot`, as it uses that only as the starting point. 

This whole interface is kind of bad, but I didn't want to break existing use cases. Properly we shouldn't be using `git` to derive the module path, but I wanted to avoid having to update all our existing config files.

internal/gomod/module.go
creachadair: Either or both.

creachadair: For further context: I think it was a mistake for us to write the indexer in such a way that it will touch the `go` tool for anything other than reading metadata. Indexing should either be done _during_ a build, so that the dependencies for each package being indexed are already populated, or _after_ a build so that the dependencies of the whole module are populated.

But that ship may have already sailed, in which case doing it here may be the least-pessimal workaround.

cmd/lsif-go/main.go
efritz: Same here. But I got the double spaces whipped out of me.

creachadair: The idealists who believe computers can automatically get spacing correct without that signal are refreshingly optimistic in their naïveté. Even Dr. Knuth realized it required input from the user, yet there are some communities you cannot blast off their high horses. (I've removed the spaces, since I really don't care that much especially in plain text comments—but I reserve the right to continue laughing and shaking my head at abject prescriptivists)

internal/gomod/module.go
creachadair: I added a call to `go mod download`. That preserves the existing behaviour for modules without vendored dependencies.

cmd/lsif-go/main.go
efritz: Would now be the item to update the interface?

efritz:
I'm assuming that this works fine on the repos we're currently indexing in CI.

efritz (approved):

creachadair:
> I'm assuming that this works fine on the repos we're currently indexing in CI.

I tested it against a few locally and it works fine, but I haven't actually tried it in GitHub Actions yet. I think I may need to update the actions repo to pick up the new version. When I do, I'll check it against some live ones and roll it back if I missed something.

PR #54, opened 2020-04-13
Ensure go.mod field exists before using download command

efritz:


internal/gomod/module.go
```diff
@@ -114,3 +121,15 @@ func run(dir, command string, args ...string) (string, error) {
 
 	return strings.TrimSpace(out.String()), nil
 }
+
+// isModule determines if there is a go.mod file in the given project root.
+func isModule(projectRoot string) (bool, error) {
```
creachadair, line 126: This is fine as you wrote it, but I would advise simplifying it on the principle that an error here doesn't matter: It's not a module if we can't read it either:

```go
_, err := os.Stat(filepath.Join(projectRoot, "go.mod"))
if err == nil || os.IsNotExist(err) {
  return true
}
log.Printf("WARNING: Error trying to locate go.mod: %v", err)
return false
```

This will then permit a simple usage above.

creachadair: Sorry, my example was bad, but I think the goal is still good:

```go
_, err := os.Stat(filepath.Join(projectRoot, "go.mod"))
if err != nil {
  if !os.IsNotExist(err) {
    log.Printf("…")
  }
  return false
}
return true
```

efritz: If we can't state it for some _other_ reason shouldn't we error? It's unlikely, but if we don't have permissions to read the repo, for example, we'll just print a warning when we should probably fail harder.

creachadair: I don't believe it's worth plumbing other errors in this context. Other reasons for failure include things like permission errors, missing directory structure, etc., all of which would have failed the workflow before we got to this point. I'm not saying it's wrong to check, only that we can't do anything useful about it except log and fail anyway, so we might as well just do that.

If you're asking "should we exit nonzero if the go.mod file can't be found?" then yes, I think that's a reasonable step to take.

efritz: Updated.

PR #63, opened 2020-05-05
Repair monikers for fields and methods

efritz:
When writing integration tests for the precise-code-intel services, I noticed that we generate non-unique monikers for definitions. This is pretty bad.

For example, zap.String is a function in uber-go/zap. Its moniker was `{package}:String`. There were some other types that conformed to the Stringer interface in the same package that were given the same moniker. This results in having FindRefs return all references to any String method/function defined in the package instead of the specific instance the user was targeting.

This PR updates method monikers to include the receiver type name, and updates fields to include the name of the type spec in which they are embedded.

### DIFF
--- a/internal/index/helper.go
+++ b/internal/index/helper.go
@@ -14,6 +14,46 @@ import (
 	"golang.org/x/tools/go/packages"
 )
 
+// monikerIdentifier constructs a moniker identifier from an ident. We can't just use the
+// ident name itself as this causes non-unique moniker generations (such as a String function
+// and a struct that conforms to the Stringer interface).
+func monikerIdentifier(f *ast.File, pkgPath string, ident *ast.Ident, obj types.Object) string {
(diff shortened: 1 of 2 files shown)


internal/index/helper.go
```diff
@@ -14,6 +14,46 @@ import (
...
+
+	return fmt.Sprintf("%s:%s", pkgPath, ident.String())
+}
+
+func safeTrim(s, prefix string) string {
+	if strings.HasPrefix(s, prefix) {
+		return s[len(prefix):]
+	}
+	return s
+}
```
unknwon, line 55: qq: what is the difference between this and `strings.TrimPrefix`?

efritz: Uh... nothing. For some reason i thought TrimPrefix would trim a matching prefix multiple times.

Will update in master.

PR #66, opened 2020-07-28
Single-pass hover text extraction

efritz:
The current implemtnation of lsif-go contains a super-quadratic way of determining comments based on token positions. For each definition, we trace the AST from the root of the containing file through nodes that intersect the target position (via `astutil.PathEnclosingInterval`) then extract the comments by tracing the constructed ancestor path backwards.

This requires us to scan the same file from the root for every definition that exists in that file. **Much** of this work will be repeated and we'll end up tracing the same path multiple times to get to definitions with slightly different paths. [Tell me that's not a bad idea in a world containing files of this size](https://github.com/aws/aws-sdk-go/blob/master/service/ec2/api.go).

This change introduces a pre-warmed cache of hover text that is extracted by traversing the AST of each file _once_. We further speed up this process by walking GOMAXPROC files in parallel.

- Before this change, aws-sdk-go indexes in 7m11s.
- After this change, aws-sdk-go indexes in 2m9s.

### DIFF
--- a/internal/index/helper.go
+++ b/internal/index/helper.go
@@ -66,9 +66,9 @@ func lspRange(pos token.Position, name string, isQuotedPkgName bool) (start prot
 }
 
 // findContents returns contents used as hover info for given object.
-func findContents(pkgs []*packages.Package, p *packages.Package, f *ast.File, obj types.Object) ([]protocol.MarkedString, error) {
+func findContents(pkgs []*packages.Package, p *packages.Package, f *ast.File, obj types.Object, hoverLoader *HoverLoader) ([]protocol.MarkedString, error) {
 	s, extra := typeString(obj)
-	comments, err := findComments(pkgs, p, f, obj)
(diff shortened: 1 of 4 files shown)


keegancsmith:
Drive by comment: I'm unsure how applicable this package is, but it might be a nice performant architecture for writing things which extract information from go. https://godoc.org/golang.org/x/tools/go/analysis

efritz:
@keegancsmith this would be useful to report diagnostics but wouldn't replace any work here.

keegancsmith:
> @keegancsmith this would be useful to report diagnostics but wouldn't replace any work here.

Should probably fork this into another discussion, but take a closer look at that package. It is a generic way to build tools which extract information from go programs in efficient ways. IE takes care of things like caching intermediate artefacts based on the build graph. However, given we don't really have a good way to do incremental work for `lsif-go` relying on `packages` makes a lot of sense.

PR #69, opened 2020-08-07
Major refactor

efritz:
This is a major code reorganization, but _should_ not have any major behavioral effects. There is increased test (coverage 80% of the indexer package) and documentation (each function).

The only change that should be visible is a nicer output (which outputs the progress of each stage of indexing).


efritz:
I am merging this now _without_ a major version bump so that I can work on some additional low-hanging performance issues (that will be reviewable) before running benchmarks against the old version.

Feel free to look through the new code to see if there are any smells that have been uncovered by the reorganization here.

PR #72, opened 2020-08-08
Refactor emitter

efritz:
This PR does a few things that allowed us to clean up a bunch of stuff in the indexer:

- Output uint64 identifiers instead of strings
- Structure the emitter/writer so that they can be used from multiple goroutines (will help us parallelize later)
- Remove unused err checks (where they are always nil). This forms a lot of the changes in the indexer pacakge.


efritz:
@aidaeology 

Subsumed by https://github.com/sourcegraph/lsif-go/pull/73 and https://github.com/sourcegraph/lsif-go/pull/74.

PR #77, opened 2020-08-09
Make monikers for struct fields more efficient

efritz:
Similar to the way we preload hover text (see https://github.com/sourcegraph/lsif-go/commit/c165bfff52e9d4f87891bba497e3b70fea144d89), we can also preload nested container names for moniker paths. This saves us from re-tracing the same AST nodes.

We piggyback on the hover text preload step and keep track of the relevant parent nodes and stash them into a map for later fast retrieval. There's still some small opportunities here for cleaner code which I'll probably tackle later (for example, we keep the entire ast path to the current node but we only look at the last three, and only look at certain node types - we can store fewer things during this traversal).

*Reviewers*: The logic changes are in the first two commits. The remaining commits are code style changes, a struct rename (HoverLoader to Preloader), and test updates.


gbrik:
which commits contain the logic changes? when i pick the two oldest it shows me 29 files changed, even though the overall PR is 10 files changed

efritz:
@gbrik Let me make a review order.

efritz:
@gbrik 

`internal/indexer/moniker.go` has the original logic for monikers, which uses astutil.PathEnclosingInterval to walk to the AST from the root of the file to the node to get a path from which to extract the names of typespecs and fields. This was moved into `internal/indexer/preloader.go`, which was renamed from HoverLoader. The file was a rename so the diff is a bit wonky, but the only relevant lines are https://github.com/sourcegraph/lsif-go/pull/77/files#diff-34daa693f818ed4b676861a3a035e975R67 and https://github.com/sourcegraph/lsif-go/pull/77/files#diff-34daa693f818ed4b676861a3a035e975R86.

The idea here is to walk each file's AST only once and construct the moniker path for every position contained in a typespec or a field. It's then a simple lookup instead of a recalculation.

All other changes are a rename of a struct.

efritz:
@gbrik Here are the two commits with the logic changes:

https://github.com/sourcegraph/lsif-go/pull/77/commits/ef56b06b108c91cd4a300e1db00bacb15a9a23cd
https://github.com/sourcegraph/lsif-go/pull/77/commits/46218484686f27d04cc4932074d6d72fa9093c20

PR #84, opened 2020-08-10
Refine positions during AST traversal to reduce search space

efritz:
We do binary search on _all_ positions for a package, which is unnecessary. As we traverse the AST we can only pass along the positions which are valid for the parent (children necessarily fall in the same bounds).

### DIFF
--- a/internal/indexer/preloader.go
+++ b/internal/indexer/preloader.go
@@ -3,6 +3,7 @@ package indexer
 import (
 	"go/ast"
 	"go/token"
+	"sort"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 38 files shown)


internal/indexer/preloader.go
```diff
@@ -83,17 +84,23 @@ func visit(
 ) {
 	newPath := updateNodePath(path, node)
 	newMonikerPath := updateMonikerPath(monikerPath, node)
+	start := sort.Search(len(positions), func(i int) bool {
+		return positions[i] >= node.Pos()
+	})
+
+	end := start
+	for end < len(positions) && positions[end] <= node.End() {
```
gbrik, line 92: perf nit since it's a perf PR: 
finding end would be faster using another application of sort.Search, but idk how many positions an ast node can span for whether it matters

efritz: My assumption is that it's fairly small and I don't want to break cache locality here.

gbrik: makes sense

PR #95, opened 2020-08-18
Fix module version inference

efritz:
The way we were asking git for tags to construct a module version was broken. We were checking to see if the repo had _any_ tags, then assumed that the commit they were indexing would be able to find it. This is not true if the tags are on a branch that's not reachable from the current commit, or the first tag comes after the current commit.

This fixes that issue by using the `--always` flag, which will return the revhash if no tags are found.

This should close https://github.com/sourcegraph/sourcegraph/issues/12487.


Strum355 (approved):
LGTM :shipit:

PR #96, opened 2020-08-20
Collapse reference result info and definition info structs

efritz:
Every place we set a DefinitionInfo we also set a ReferenceResultInfo. This contains some of the same information (result set identifier and the document/range identifiers that define it). This collapses this into one structure so we don't need to double-allocate heap space for this redundant information.

### DIFF
--- a/internal/indexer/indexer.go
+++ b/internal/indexer/indexer.go
@@ -27,22 +27,21 @@ type Indexer struct {
 	silent         bool              // Whether to suppress all output
 
 	// Definition type cache
-	consts  map[token.Pos]*DefinitionInfo // position -> info
-	funcs   map[string]*DefinitionInfo    // name -> info
-	imports map[token.Pos]*DefinitionInfo // position -> info
-	labels  map[token.Pos]*DefinitionInfo // position -> info
-	types   map[string]*DefinitionInfo    // name -> info
-	vars    map[token.Pos]*DefinitionInfo // position -> info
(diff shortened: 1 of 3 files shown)


internal/indexer/indexer.go
```diff
@@ -27,22 +27,21 @@ type Indexer struct {
...
-	imports map[token.Pos]*DefinitionInfo // position -> info
-	labels  map[token.Pos]*DefinitionInfo // position -> info
-	types   map[string]*DefinitionInfo    // name -> info
-	vars    map[token.Pos]*DefinitionInfo // position -> info
+	consts  map[interface{}]*DefinitionInfo // position -> info
+	funcs   map[interface{}]*DefinitionInfo // name -> info
+	imports map[interface{}]*DefinitionInfo // position -> info
+	labels  map[interface{}]*DefinitionInfo // position -> info
+	types   map[interface{}]*DefinitionInfo // name -> info
+	vars    map[interface{}]*DefinitionInfo // position -> info
```
Strum355, line 35: If we move to golang:1.15, we can probably reap the benefits that it brought along for int types stored in interface{} [Converting a small integer value into an interface value no longer causes allocation.](https://golang.org/doc/go1.15#runtime). Hopefully that also applies for token.Pos

efritz: https://github.com/sourcegraph/lsif-go/pull/97

Strum355 (approved):
:shipit:

PR #99, opened 2020-08-20
De-pointerify protocols structures

efritz:
Use value semantics wherever possible for smaller, immutable structures like these.


Strum355:
Just so I'm sure, how did you determine that this reduces heap allocations? As far as I'm aware, Go's escape analysis doesn't tend to be "*the best*", and that in many of these functions, given the return value is passed to `e.writer.Write(...)` which takes an interface{} as its param, they are hot candidates for being escaped to the heap just for being passed through interface{}. `-gcflags="-m"` should tell you at build time if they escape or not, dont know if you were aware of it : )

efritz:
Anything that is sent on a channel escapes, and all these values are sent on the writer channel.

efritz:
This change isn't necessarily 100% due to performance - there's no reason for these to be pointers in the first place and it's a hold-over from the original draft.

Strum355 (approved):

PR #100, opened 2020-08-20
Remove phase duration

efritz:
This is left-over debugging data.


Strum355 (approved):
If this is info we don't want anymore (or already have some place else), then this LGTM

PR #104, opened 2020-08-21
Preload on demand

efritz:
This updates the preloader to only load data for packages on their first use. This has the effect of only loading packages just-in-time, which saves us the pre-step of parsing all packages at once. This may also save some memory if packages are imported but no identifier is referenced in a way in which we need to extract hover text or get a moniker path.


Strum355 (approved):
Dont see any glaring issues, LGTM :)

PR #108, opened 2020-08-24
Use jsoniter for marshalling everywhere

efritz:


Strum355:
Is https://github.com/sourcegraph/lsif-go/blob/master/internal/indexer/protocol_test.go gonna be moved to json-iterator too or is it fine as-is?

efritz:
> Is https://github.com/sourcegraph/lsif-go/blob/master/internal/indexer/protocol_test.go gonna be moved to json-iterator too or is it fine as-is?

I think it's fine as is. I just wanted to move the stuff in the actual encoding path to this because it relies less heavily on reflection.

Strum355 (approved):

PR #119, opened 2020-09-03
Do not output an unreferenced reference result

efritz:
This results in slightly smaller dumps.


gbrik (approved):

gbrik:
wait oops i spoke too soon!

gbrik:
i think this means that if you find-refs on an unused symbol, it will return no ref results, instead of a single ref result being that symbol

efritz:
I believe this only affects un-referencable things (like `_`) but I'll try to repro.

efritz:
I just tried an upload for lsif-go looking at the gomod Download function, which is no longer referenced. It had one semantic reference result which was itself.

efritz:
Ah looks like you're right. I'm going to revert this and see if there's some additional way to determine when this is useless.

gbrik:
sry for the itchy approve finger

efritz:
Reverted!

Issue #117, opened 2020-09-03
bug: no J2D or find-refs of switch-case blocks over an object's type


gbrik:
Code like the following:
```
switch v := obj.(type) {
  case A: f1(v, v)
  case B: f2(v, v)
}
```
does not get indexed properly because when processing definitions, the definition for `v` has a `nil` associated def object, since there's a different def object for the v in each branch. Then when we encounter each use of `v`, there is an associated def object but since it's not indexed in our definitions we only output a hover result.

Discussion:
- If there's a way to know that this is why the def object is `nil` during the definitions pass, that would likely be the right way to solve.
- What LSIF should we actually output here? IMO:
  - Each case branch should have a different hover result cause they have different types. 
  - The def result is canonical.
  - I think the reference result should point to all uses in every branch. Kind of weird to have a reference return results with different types, but I think the experience would be weirder if you hover over one of these and only the symbols in that branch highlight. Also the definition would need to have everything as a reference result anyway.
  - Open question: what should go in the hover result when hovering the definition?

Therefore the resulting LSIF structure should be:

```
vdef -> resultSet1
resultSet1 -> hoverResult1
resultSet1 -> resultSetShared
resultSetShared -> defResult1
resultSetShared -> refResult1
defResult1 -> vdef
refResult1 -> vdef
vref1 -> resultSet2
resultSet2 -> hoverResult2
resultSet2 -> resultSetShared
refResult1 -> vref1
vref2 -> resuiltSet2
refResult1 -> vref2
vref3 -> resultSet3
resultSet3 -> hoverResult3
resultSet3 -> resultSetShared
refResult1 -> vref3
vref4 -> resuiltSet3
refResult1 -> vref4
```

Thoughts?

Also how does the worker process this graph?

efritz:
Looking into this found some curious comments that are influencing this data to be missing in the graph:

```go
type types.Info struct {
	// Defs maps identifiers to the objects they define (including
	// package names, dots "." of dot-imports, and blank "_" identifiers).
	// For identifiers that do not denote objects (e.g., the package name
	// in package clauses, or symbolic variables t in t := x.(type) of
	// type switch headers), the corresponding objects are nil.
	//
	// For an embedded field, Defs returns the field *Var it defines.
	//
	// Invariant: Defs[id] == nil || Defs[id].Pos() == id.Pos()
	Defs map[*ast.Ident]Object

	// Implicits maps nodes to their implicitly declared objects, if any.
[...]

PR #121, opened 2020-09-04
Remove uses of ast.Idents

efritz:
The only information we have is also attached to the objects, which allows us to treat things more uniformly. This is the first step into making the solution for https://github.com/sourcegraph/lsif-go/issues/117 not messy.

### DIFF
--- a/internal/indexer/indexer.go
+++ b/internal/indexer/indexer.go
@@ -247,7 +247,7 @@ func (i *Indexer) indexDefinitions() {
 
 // indexDefinitionsForPackage emits data for each definition within the given package.
 func (i *Indexer) indexDefinitionsForPackage(p *packages.Package) {
-	for ident, obj := range p.TypesInfo.Defs {
+	for _, obj := range p.TypesInfo.Defs {
 		if obj == nil {
 			continue
 		}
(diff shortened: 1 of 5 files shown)


internal/indexer/indexer.go
```diff
@@ -391,13 +391,17 @@ func (i *Indexer) indexReferences() {
 
 // indexReferencesForPackage emits data for each reference within the given package.
 func (i *Indexer) indexReferencesForPackage(p *packages.Package) {
-	for ident, obj := range p.TypesInfo.Uses {
-		pos, d, ok := i.positionAndDocument(p, ident, obj)
+	for ident, definitionObj := range p.TypesInfo.Uses {
+		if definitionObj == nil {
+			continue
+		}
```
gbrik, line 397: question:
just for understanding `golang.org/x/tools/go/packages`'s sake, in what cases does this happen?

gbrik (approved):
yum, LGTM! (the G stands for gourmet)

internal/indexer/indexer.go
efritz: I don't know - I may have seen it on `TypeInfo.Uses` only but I'd rather be safe here since I don't know why it would be nil in either of these places 😬

efritz: I'll look into it though and post here if I'm enlightened.

efritz: For `TypeInfo.Defs`:

> For identifiers that do not denote objects (e.g., the package name in package clauses, or symbolic variables t in t := x.(type) of type switch headers), the corresponding objects are nil.

PR #122, opened 2020-09-04
Index symbolic variable in type switch header

efritz:
**Previously**: There is no definition for the symbolic variable in the type switch header, and each of the case clauses share _some_ arbitrary hover text that applies to one case clause correctly, but all the remaining clauses incorrectly. This is also a non-deterministic choice and may change if you index twice.

Kapture 2020-09-04 at 12 17 22

**After this change**: We have the correct hover text for each case clause use, and each occurrence of the symbol variable are linked correctly via def/ref chains. The definition itself has no hover text (see TODO in PR code).

Kapture 2020-09-04 at 12 18 09

Closes https://github.com/sourcegraph/lsif-go/issues/117.

**Remaining work:**

- [ ] Figure out how to get hover text for definition
- [x] Write additional tests covering type switches

### DIFF
--- a/internal/indexer/indexer.go
+++ b/internal/indexer/indexer.go
@@ -247,9 +247,31 @@ func (i *Indexer) indexDefinitions() {
 
 // indexDefinitionsForPackage emits data for each definition within the given package.
 func (i *Indexer) indexDefinitionsForPackage(p *packages.Package) {
-	for _, obj := range p.TypesInfo.Defs {
+	// Create a map of implicit case clause objects by their position. Note that there is an
+	// implicit object for each case clause of a type switch (including default), and they all
+	// share the same position. This creates a map with one arbitrarily chosen argument for
+	// each distinct type switch.
+	caseClauses := map[token.Pos]types.Object{}
+	for node, obj := range p.TypesInfo.Implicits {
+		if _, ok := node.(*ast.CaseClause); ok {
+			caseClauses[obj.Pos()] = obj
+		}
+	}
+
+	for ident, obj := range p.TypesInfo.Defs {
+		typeSwitchHeader := false
 		if obj == nil {
-			continue
+			// The definitions map contains nil objects for symbolic variables t in t := x.(type)
+			// of type switch headers. In these cases we select an arbitrary case clause for the
+			// same type switch to index the definition. We mark this object as a typeSwitchHeader
+			// so that it can distinguished from other definitions with non-nil objects.
+			caseClause, ok := caseClauses[ident.Pos()]
+			if !ok {
+				continue
+			}
+
+			obj = caseClause
+			typeSwitchHeader = true
 		}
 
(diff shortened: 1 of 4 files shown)


aidaeology (approved):

internal/indexer/indexer.go
```diff
@@ -445,14 +473,24 @@ func (i *Indexer) getDefinitionInfo(obj types.Object) *DefinitionInfo {
...
+func (i *Indexer) indexReferenceToDefinition(p *packages.Package, document *DocumentInfo, pos token.Position, definitionObj types.Object, d *DefinitionInfo) (uint64, bool) {
 	rangeID := i.ensureRangeFor(pos, definitionObj)
 	_ = i.emitter.EmitNext(rangeID, d.ResultSetID)
 
 	d.m.Lock()
 	d.ReferenceRangeIDs[document.DocumentID] = append(d.ReferenceRangeIDs[document.DocumentID], rangeID)
 	d.m.Unlock()
 
+	if d.TypeSwitchHeader {
+		// Attache a hover text result _directly_ to the given range so that it "overwrites" the
```
Strum355, line 485: s/Attache/attach/

Strum355 (approved):

internal/indexer/indexer.go
efritz: No it's fancy.

internal/indexer/indexer.go
```diff
@@ -309,22 +331,27 @@ func (i *Indexer) markRange(pos token.Position) bool {
...
 
 	_ = i.emitter.EmitNext(rangeID, resultSetID)
 	_ = i.emitter.EmitTextDocumentDefinition(resultSetID, defResultID)
 	_ = i.emitter.EmitItem(defResultID, []uint64{rangeID}, document.DocumentID)
-	_ = i.emitter.EmitTextDocumentHover(resultSetID, hoverResultID)
+
+	if typeSwitchHeader {
+		// TODO(efritz) - not sure how to document a type switch header symbolic variable
+		// I'd like to somehow keep the type string of the RHS, but I'm not yet sure how
+		// to resolve that data given what we have.
```
Strum355, line 346: if you could give an example of what youre thinking here and how you envisage it being displayed to the user thatd be really appreciated so I've an idea whats going on :)

Strum355 (approved):

internal/indexer/indexer.go
efritz: I'm not sure yet, but I'd think something like:

```go
var v interface{} = 30

switch e := v.(type) {
// ...
}
```

Then e's hover text would be `var e interface{}` in the header.

efritz:
I'm going to merge this now since it's a huge improvement, despite the remaining TODO note. This will take a bit more work but isn't a blocker from shipping the fix as it is currently.

PR #125, opened 2020-10-19
fix docs link

phanirithvij:


Strum355 (approved):
Thanks for the fix! :tada:

PR #129, opened 2021-01-29
Symbols with textDocument/documentSymbol

beyang:


internal/indexer/indexer.go
```diff
@@ -337,30 +372,9 @@ func (i *Indexer) positionAndDocument(p *packages.Package, pos token.Pos) (token
 	return position, d, true
 }
 
-// markRange sets an empty range identifier in the ranges map for the given position.
-// If a  range for this identifier has already been marked, this method returns false.
```
beyang, line 341: @efritz It looks like the point of this method was to prevent multiple definitions/references associated with the same range. With the addition of symbols, we now expect most definitions' ranges to *also* be associated with a symbol, which means those ranges will be emitted in the `indexSymbols` stage. Therefore, this method is no longer valid (keeping it as is would mean we don't emit definitions whose ranges have already been emitted in the symbols stage). Deleting it didn't break any tests. Let me know if there should still be a check that enforces non-duplication for some subset of range-adjacent vertices emitted.

internal/indexer/indexer.go
```diff
@@ -337,30 +372,9 @@ func (i *Indexer) positionAndDocument(p *packages.Package, pos token.Pos) (token
...
-	}
-
-	i.ranges[pos.Filename][pos.Offset] = 0 // placeholder
-	return true
-}
-
 // indexDefinition emits data for the given definition object.
 func (i *Indexer) indexDefinition(p *packages.Package, filename string, document *DocumentInfo, pos token.Position, obj types.Object, typeSwitchHeader bool, ident *ast.Ident) uint64 {
-	rangeID := i.emitter.EmitRange(rangeForObject(obj, pos))
+	rangeID := i.ensureRangeFor(pos, obj)
```
beyang, line 377: Because the range may have already been emitted in the `indexSymbols` stage.

internal/indexer/symbols.go
```diff
@@ -0,0 +1,196 @@
...
+// https://microsoft.github.io/language-server-protocol/specifications/lsif/0.5.0/specification/#documentSymbol.
+//
+// This data includes:
+// - the textDocument/documentSymbol edge
+// - the documentSymbolResult vertex, which contains a hierarchy of RangeBasedDocumentSymbol vertices
+// - ranges associated with each symbol, with the appropriate range tag
+//
+// Because the textDocument/documentSymbol spec is document-based, all symbol hierarchies are rooted
+// at documents. Top-level variables, functions, types, constants, etc. are children of the root,
+// and methods are children of their receiver type.
```
beyang, line 31: This is the main method of this PR. Probably makes sense to start here.

go.mod
```diff
@@ -15,3 +15,5 @@ require (
 	github.com/sourcegraph/lsif-protocol v1.0.0
 	golang.org/x/tools v0.1.0
 )
+
+replace github.com/sourcegraph/lsif-protocol v1.0.0 => ../lsif-protocol
```
beyang, line 19: TODO(beyang): update after https://github.com/sourcegraph/lsif-protocol/pull/6 is merged.

internal/testdata/child_symbols.go
```diff
@@ -0,0 +1,12 @@
+// Package testdata
+package testdata
+
+type Struct struct {
+	Field string
```
efritz, line 5: You may also want to have nested types here as a test case, e.g.

```
type Struct struct {
	Anonymous struct {
		FieldA int
		FieldB int
		FieldC int
	}
}
```

internal/indexer/indexer.go
```diff
@@ -574,11 +588,37 @@ func (i *Indexer) ensureRangeFor(pos token.Position, obj types.Object) uint64 {
...
+	defer i.stripedMutex.UnlockKey(pos.Filename)
+
+	line := pos.Line - 1
+	column := pos.Column - 1
+
+	start := protocol.Pos{Line: line, Character: column}
+	end := protocol.Pos{Line: line, Character: column + length}
+
+	if rangeID, ok := i.ranges[pos.Filename][pos.Offset]; ok {
+		log.Printf("Duplicate range already exists at %s:%d, tag may be absent", pos.Filename, pos.Offset)
```
efritz, line 607: Can you guard this output like we do with the rest? There's verbosity levels we can use here (and there's also output animation that rogue prints will mess up).

internal/indexer/indexer.go
efritz: I have to think a little harder about this, but will definitely help testing on some examples to make sure there's no big performance/space regression here.

internal/indexer/indexer.go
```diff
@@ -574,11 +588,37 @@ func (i *Indexer) ensureRangeFor(pos token.Position, obj types.Object) uint64 {
 		return rangeID
 	}
 
-	rangeID = i.emitter.EmitRange(start, end)
+	rangeID = i.emitter.EmitRange(start, end, nil)
```
efritz, line 591: Could simplify this based on comments in other pr.

internal/indexer/indexer.go
```diff
@@ -142,6 +144,39 @@ func (i *Indexer) loadPackages() error {
 		i.packages = pkgs
 		i.packagesByFile = map[string][]*packages.Package{}
 
+		shouldVisitPackage := func(p *packages.Package) bool {
```
efritz, line 147: Oh cool. Could we extract this into a method on the indexer and add a few unit tests around it?

internal/indexer/symbols.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+	}
+
+	// Protect with mutex, because multiple packages with the same package path can share files and
+	// doc.NewFromFiles will modify the elements of the file slice.
+	i.stripedMutex.LockKey(p.PkgPath)
+	docpkg, err := doc.NewFromFiles(p.Fset, files, p.PkgPath, doc.AllDecls|doc.PreserveAST)
+	if err != nil {
+		i.stripedMutex.UnlockKey(p.PkgPath)
+		fmt.Fprintf(os.Stderr, "Package %s aborted: doc.NewFromFiles failed with error: %s", p.PkgPath, err)
```
efritz, line 51: This should likely return an error to the caller to check instead of just emitting an error and silently passing here.

internal/indexer/symbols.go
```diff
@@ -0,0 +1,196 @@
...
+		return
+	}
+	i.stripedMutex.UnlockKey(p.PkgPath)
+
+	for _, file := range files {
+		filename := p.Fset.Position(file.Pos()).Filename
+		if _, ok := i.documents[filename]; !ok {
+			log.Printf("Skipping file symbol because file not found: %s", filename)
```
efritz, line 59: Same comment here about logs.

internal/indexer/symbols.go
```diff
@@ -0,0 +1,196 @@
...
+				},
+			}, nil)
+		}
+	}
+
+	// Index result vertices and edges
+	for filename, symbolData := range documentSymbols {
+		doc := i.documents[filename]
+		if doc == nil {
+			log.Printf("Error: document not found, skipping (filename: %s)", filename)
```
efritz, line 162: Same

internal/indexer/indexer.go
efritz: I think this could be useful outside of this PR - it might be worth breaking this out and merging it earlier.

efritz (changes requested):
A few questions/comments inline.

internal/indexer/indexer.go
efritz: I'm unsure why you need multiple definitions to a range - is this because you're nesting tags inside of the range vertex itself?

tjdevries (commented):
I'll have more questions :)

internal/indexer/symbols.go
```diff
@@ -0,0 +1,196 @@
...
+	// Protect with mutex, because multiple packages with the same package path can share files and
+	// doc.NewFromFiles will modify the elements of the file slice.
+	i.stripedMutex.LockKey(p.PkgPath)
+	docpkg, err := doc.NewFromFiles(p.Fset, files, p.PkgPath, doc.AllDecls|doc.PreserveAST)
+	if err != nil {
+		i.stripedMutex.UnlockKey(p.PkgPath)
+		fmt.Fprintf(os.Stderr, "Package %s aborted: doc.NewFromFiles failed with error: %s", p.PkgPath, err)
+		return
+	}
+	i.stripedMutex.UnlockKey(p.PkgPath)
```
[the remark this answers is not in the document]
efritz: You need to do it manually instead of using a defer in order to (1) ensure it's unlocked and (2) keep the critical section guarded by the lock small. You'll see this idiom frequently around locks where one exit is a conditional early-exit.

internal/indexer/indexer.go
slimsag: I'm looking at doing this as a first PR.

internal/indexer/indexer.go
```diff
@@ -142,6 +144,39 @@ func (i *Indexer) loadPackages() error {
...
+			// want to avoid duplication.
+			if p.Name == "main" && strings.HasSuffix(p.ID, ".test") {
+				return false // synthesized `go test` program
+			}
+			if strings.HasSuffix(p.Name, "_test") {
+				return true
+			}
+
+			// Index only the combined test package if it's present. If the package has no test files,
+			// it won't be present, and we need to just index the default package.
```
slimsag, line 158: @efritz @beyang can either of you explain to me what "index only the combined test package if it's present." means? I am specifically confused how two packages can be _combined_? Is this an LSIF concept, or a `go/loader` concept I'm not aware of?

internal/indexer/indexer.go
slimsag: @beyang can you answer "I'm unsure why you need multiple definitions to a range - is this because you're nesting tags inside of the range vertex itself?"

slimsag: (I want to make sure I understand)

internal/indexer/indexer.go
slimsag: Figured it out: https://github.com/sourcegraph/lsif-go/pull/138/files#diff-87c2475baa87005bd3a3001ec33f201feea23caddbbf5a2f249238a2bd32026dR158-R170

internal/indexer/symbols.go
```diff
@@ -0,0 +1,184 @@
+package indexer
+
+import (
+	"fmt"
+	"go/ast"
+	"go/doc"
```
slimsag, line 6: There seems to be some historical precedent for why we should not use go/doc for generating document symbols from the go-langserver days: https://github.com/sourcegraph/go-langserver/commit/00cb77287a8d0f185e5cb978a7c75197609b9f54#diff-eb4c6c7112f00549c81d7f1a012322112eaf0d38f3d46035e89a8b323779923b

I can also imagine the go/doc interface becoming a limiting factor in the future i we try to support bi-directional linking of symbols, such as when Go generics arrive. I will look into whether or not it makes sense to reuse what we've done historically in go-langserver instead.

internal/indexer/indexer.go
slimsag: answers: https://github.com/sourcegraph/lsif-go/pull/142

internal/indexer/indexer.go
slimsag: answers: https://github.com/sourcegraph/lsif-go/pull/142

slimsag:
I am taking a fairly different approach for several reasons, and have begun merging other PRs which aim to support this. You can find them all linked in https://github.com/sourcegraph/sourcegraph/issues/19591

Closing / discarding this.

PR #131, opened 2021-02-10
fix type aliases associated with wrong definition

### LINKED ISSUE #130: cross-package type alias definitions have self-referencing semantic definition results
Go To Definition on `sfnt.Font` in the definition of `opentype.Font` at https://sourcegraph.com/github.com/golang/image@35266b937fa6/-/blob/font/opentype/opentype.go#L67:6 goes to `opentype.Font`. I expected it to instead take me to the definition of `sfnt.Font`, which is in a different package in the same module.

Strum355:
closes #130 

There may be an issue if there are multiple idents in the index with the same name that alias the same type. Should we refine the conflict resolution?


efritz (changes requested):
Can you add in your test case?

Strum355:
Knew I forgot something :facepalm:

efritz (approved):
secret 🍔

Issue #133, opened 2021-02-24
Processor failing during CI/CD


emchap:
Customer reported as well, filed separately here: https://github.com/sourcegraph/customer/issues/216

The processor is failing during CI/CD; log are linked in the customer issue for privacy. Reverting to 1.2 resolves the issue. Top error:

`panic: interface conversion: types.Type is *types.Struct, not *types.Named`

@Strum355 reported that he found a public repo demonstrating the issue as well.

Strum355:
Issue has been narrowed down to cases of type aliases to unnamed struct eg
```go
type Secret = struct { ... }
```
All personal gripes about this aside, this should be fast enough to fix. Thanks for the report :slightly_smiling_face:

PR #134, opened 2021-02-24
Fix assertion panic for type aliases to anonymous structs

efritz:
Fixes https://github.com/sourcegraph/lsif-go/issues/133

### DIFF
--- a/internal/indexer/indexer_test.go
+++ b/internal/indexer/indexer_test.go
@@ -240,49 +240,86 @@ func TestIndexer(t *testing.T) {
 			if value := strings.ReplaceAll(hover.Result.Contents[2].Value, "  ", "\t"); value != expectedUnderlyingType {
 				t.Errorf("incorrect hover text documentation. want=%q have=%q", expectedUnderlyingType, value)
 			}
+		}
 
-			{
-				r, ok := findRange(w.elements, typealiasFile, 7, 27)
-				if !ok {
-					t.Fatalf("could not find target range")
-				}
+		{
+			r, ok := findRange(w.elements, typealiasFile, 7, 27)
+			if !ok {
+				t.Fatalf("could not find target range")
+			}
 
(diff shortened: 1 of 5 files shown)


internal/indexer/indexer_test.go
```diff
@@ -240,49 +240,86 @@ func TestIndexer(t *testing.T) {
 			if value := strings.ReplaceAll(hover.Result.Contents[2].Value, "  ", "\t"); value != expectedUnderlyingType {
 				t.Errorf("incorrect hover text documentation. want=%q have=%q", expectedUnderlyingType, value)
 			}
+		}
```
efritz, line 243: There's actually 3 tests here, should we break these out into different t.Runs?

internal/indexer/typestring_test.go
```diff
@@ -91,6 +91,24 @@ func TestTypeStringEmptyStruct(t *testing.T) {
 	}
 }
 
+func TestTypeStringNameEqualsAnonymousStruct(t *testing.T) {
+	_, f := findDefinitionByName(t, getTestPackages(t), "TestEqualsStruct")
+
+	signature, extra := typeString(f)
+	if signature != "type TestEqualsStruct = struct" {
+		t.Errorf("unexpected type string. want=%q have=%q", "type TestEqualsStruct struct", signature)
```
efritz, line 99: This should fail?

efritz, line 99: Should this fail? I'd expect this to hit typestring.go:50

Strum355: The tests passed :thinking:

efritz: Oh the text is wrong here. I'll fix.

Strum355: Ohh yes cool cool

PR #135, opened 2021-02-24
Fix moniker of embedded structs

efritz:
Fix moniker of an anonymous struct or interface nested inside another.


internal/indexer/package_data_cache.go
```diff
@@ -169,12 +169,17 @@ func visit(
 func updateMonikerPath(monikerPath []string, node ast.Node) []string {
 	switch q := node.(type) {
 	case *ast.Field:
+		// Handle field name/names
 		if len(q.Names) > 0 {
-			// Add names of distinct fields whose type is an anonymous struct type
-			// containing the target field (e.g. `X struct { target string }`).
+			// Handles things like `a, b, c T`
```
efritz, line 174:
```suggestion
			// Handle things like `a, b, c T`. If there are multiple names we just default to the first
			// one as each field must belong on at most one moniker path. This is sub-optimal and
			// should be addressed in https://github.com/sourcegraph/lsif-go/issues/154.
```

internal/indexer/helpers_test.go
```diff
@@ -44,7 +46,16 @@ func getTestPackages(t *testing.T) []*packages.Package {
...
 func findDefinitionByName(t *testing.T, packages []*packages.Package, name string) (*packages.Package, types.Object) {
 	for _, p := range packages {
-		for _, def := range p.TypesInfo.Defs {
+		idents := make([]*ast.Ident, 0, len(p.TypesInfo.Defs))
+		for ident := range p.TypesInfo.Defs {
+			idents = append(idents, ident)
+		}
+		sort.Slice(idents, func(i, j int) bool {
+			return idents[i].Pos() < idents[j].Pos()
+		})
```
Strum355, line 55: Whats the sort needed for?

efritz: Stupid tests look at names but anonymous fields make things conflict. This sorts it so that we deterministically fetch the first one we encounter lexically in tests.

Strum355 (approved):

PR #138, opened 2021-04-01
internal/indexer: add method for determining whether a package should be indexed

slimsag:
This is pulled out of https://github.com/sourcegraph/lsif-go/pull/129#discussion_r580296179 - but with actual tests for it and more documentation explaining what it's doing and why.

Helps https://github.com/sourcegraph/sourcegraph/issues/19591


internal/indexer/indexer_test.go
```diff
@@ -336,3 +337,44 @@ func compareRange(t *testing.T, r protocol.Range, startLine, startCharacter, end
...
+	visited := map[string]bool{}
+	for _, pkg := range indexer.packages {
+		shortID := strings.Replace(pkg.ID, "github.com/sourcegraph/lsif-go/internal/testdata/internal", "…", -1)
+		if indexer.shouldVisitPackage(pkg) {
+			visited[shortID] = true
+		} else {
+			visited[shortID] = false
+		}
+	}
+	autogold.Want("visited", map[string]bool{
```
slimsag, line 369: I pulled in my [autogold](https://github.com/hexops/autogold) project here. I can imagine this being super useful in other places of this codebase - but I'm happy to do whatever here, just let me know what your preference is.

efritz: I have no issue adopting autogold + similar tools (things here are just out of past convenience). If you want to update any tests to use snapshots feel free to add additional or replace old coverage.

efritz (approved):
THANK YOU for that comment. That and the new test layout makes perfect sense now. Approving now as it looks like it's just a function addition and isn't actually used yet. When we actually apply this I can run the codeintel backend integration tests to make sure everything is still good end to end.

slimsag:
> When we actually apply this I can run the codeintel backend integration tests to make sure everything is still good end to end.

Could you explain how I myself might do that? Want to make sure I get a good picture for how to make stable changes here etc.

efritz:
Of course! I don't want to hoard the integration tests for myself. They're what are run in the codeintel-qa pipeline and instructions are found [here](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/cmd/precise-code-intel-tester/README.md).

I would re-install lsif-go on my path and then run `./scripts/clone-and-index.sh`, which will re-clone the test repos and run lsif-go in the root of each one (`./scripts/down.load.sh` will pull pre-generated indexes from GCS). Then the binary built in this dir can be used to upload and subsequently query known values within that network of indexes.

I'd also be interested to know if there's a size reduction after this PR. If the same index for this repo at master and this commit are different please comment on that (I also use an index of the [AWS SDK](https://github.com/aws/aws-sdk-go) often for something that's sizable).

slimsag:
1. I pushed a commit to actually use this function.
2. I confirmed this basically has no effect on file size in practice: I think this is because for all the data we emit already it is implicitly deduplicated by being a graph.
3. I ran the latest revision through the backend integration test suite, all good:

```
$ go build && ./precise-code-intel-tester query
All queries completed in 25.635745014s
```

For file sizes, I compared file sizes in the sourcegraph main repo:

```
$ du -sh *.lsif
240M	dump-after.lsif
240M	dump-before.lsif
```

and in the aws-go-sdk repo:

```
$ du -sh *.lsif
1.6G	dump-after.lsif
1.6G	dump-before.lsif
```

I did confirm in both cases that it *is* actually eliminating the duplicate packages - it just doesn't affect the data size we materially generate for these repos at all (but would for `textDocument/documentSymbols` I believe)

slimsag:
Gonna merge, will follow up on any future comments/suggestions.

PR #140, opened 2021-04-08
[WIP] implement textDocument/documentSymbols

slimsag:


internal/indexer/symbols.go
```diff
@@ -0,0 +1,225 @@
...
+	sym := &protocol.RangeBasedDocumentSymbol{}
+	// If the below is changed to 2, all tests pass. If it's 3, then:
+	//
+	// --- FAIL: TestIndexer (1.21s)
+	//     --- FAIL: TestIndexer/check_wg_references (0.00s)
+	//         indexer_test.go:108: incorrect reference count. want=4 have=1
+	// FAIL
+	// FAIL	github.com/sourcegraph/lsif-go/internal/indexer	18.149s
+	// FAIL
+	//
```
slimsag, line 158: Could use help understanding what is going wrong here; not sure why this causes an implicit change in behavior in reference generation

slimsag: This turned out to be due to https://github.com/sourcegraph/lsif-go/pull/142

PR #142, opened 2021-04-08
 indexer: prepare to index documentSymbols (do not index duplicate definitions)

slimsag:
After spending all day debugging, I think I finally understand what is going on here and answered my own issue :) PTAL @efritz 

Helps https://github.com/sourcegraph/sourcegraph/issues/19591

## The first commit

This prepares the indexer to index `textDocument/documentSymbols`. The problem
here is that indexing symbols involves emitting ranges for the symbol definitions,
and so does our existing indexing for `indexDefinition`. LSIF ranges cannot overlap
or be duplicative, so we must effectively make symbol and definition indexing work
together. The idea is that symbol indexing will emit ranges for the definitions it
encounters, and traditional definition indexing will merely reuse those or emit a
new range if needed.

At present, this causes all tests to fail due to the way we pre-emptively mark
a range for generation during definition indexing:

```
--- FAIL: TestIndexer (0.63s)
    --- FAIL: TestIndexer/check_Parallel_function_hover_text (0.00s)
        indexer_test.go:35: could not find target range
    --- FAIL: TestIndexer/check_external_package_hover_text (0.00s)
        indexer_test.go:62: could not find target range
    --- FAIL: TestIndexer/check_errs_definition (0.00s)
        indexer_test.go:94: incorrect definition count. want=1 have=0
    --- FAIL: TestIndexer/check_wg_references (0.00s)
        indexer_test.go:108: incorrect reference count. want=4 have=3
    --- FAIL: TestIndexer/check_NestedB_monikers (0.00s)
        indexer_test.go:122: could not find target range
    --- FAIL: TestIndexer/check_typeswitch (0.00s)
        indexer_test.go:143: could not find target range
    --- FAIL: TestIndexer/check_typealias (0.00s)
        indexer_test.go:209: could not find target range
    --- FAIL: TestIndexer/check_typealias_reference (0.00s)
        indexer_test.go:255: incorrection definition count. want=1 have=0
    --- FAIL: TestIndexer/check_typealias_anonymous_struct (0.00s)
        indexer_test.go:296: could not find target range
FAIL
FAIL	github.com/sourcegraph/lsif-go/internal/indexer	17.393s
FAIL
```

## The second commit

indexer: remove pre-emptive range marking for definitions

Prior to this change, we would pre-emptively "mark" a range for a definition
as being actively produced. This appears to have been because `indexDefinition`
would _always_ emit a new range, and this it was necessary to ensure that two
parallel definition indexing processes would not emit the same range.

Since my prior commit, however, `indexDefinition` only emits a new range if one
does not exist. So pre-emptive range marking is not required anymore.

This causes all tests to pass.

### DIFF
--- a/internal/indexer/indexer.go
+++ b/internal/indexer/indexer.go
@@ -355,7 +355,7 @@ func (i *Indexer) indexDefinitionsForPackage(p *packages.Package) {
 		}
 
 		pos, d, ok := i.positionAndDocument(p, obj.Pos())
-		if !ok || !i.markRange(pos) {
+		if !ok {
 			continue
 		}
 
@@ -387,27 +387,6 @@ func (i *Indexer) positionAndDocument(p *packages.Package, pos token.Pos) (token
 	}
 
 	return position, d, true
-}
-
-// markRange sets an empty range identifier in the ranges map for the given position.
-// If a  range for this identifier has already been marked, this method returns false.
-func (i *Indexer) markRange(pos token.Position) bool {
-	i.stripedMutex.RLockKey(pos.Filename)
-	_, ok := i.ranges[pos.Filename][pos.Offset]
-	i.stripedMutex.RUnlockKey(pos.Filename)
-	if ok {
-		return false
-	}
-
-	i.stripedMutex.LockKey(pos.Filename)
-	defer i.stripedMutex.UnlockKey(pos.Filename)
-
-	if _, ok := i.ranges[pos.Filename][pos.Offset]; ok {
-		return false
-	}
-
-	i.ranges[pos.Filename][pos.Offset] = 0 // placeholder
-	return true
 }
 
 // indexDefinition emits data for the given definition object.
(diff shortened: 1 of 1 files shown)


efritz:
I'm _becoming_ convinced this is correct now. I do worry that there could still be projects that will produce multiple _invocations_ of `indexDefinition` for the same range (which may attach a second result set).

I'm going to run this on a few projects and hand-analyze some of the output to make sure it's ok to delete that block. If something _is_ weird I think it would be an easy fix (we need distinguish a range created from symbols vs definitions, and no-op only on duplicates of the latter).

slimsag:
Sounds good! I agree that could be a concern, I didn't think of that.

slimsag:
Confirmed this also fixed the issue I was experiencing in https://github.com/sourcegraph/lsif-go/pull/140#discussion_r609162918

slimsag:
Superseded by https://github.com/sourcegraph/lsif-go/pull/143

efritz:
@slimsag Merged into here; PR should now be complete and ready for your next step.

efritz (approved):

slimsag:
Aha, thanks!

PR #143, opened 2021-04-12
Make markRange unobtrusive for document symbols

efritz:
Ok so you were correct about what `markRange` was doing. It stops multiple definitions that are reachable from different packages from clobbering over each other while trying to make the canonical version of that range.

`ensureRangeFor` was an obvious substitute since it is also properly locking so that nothing can be double-assigned. However, that lock only protects the range vertex itself, and not some other relations that should have a certain uniqueness property.

Removing `markRange` totally will make it so that ranges get `n` result sets (directly attached) to the same unique range. Thus, there are possibly `n` ambiguous definition and result sets to choose from at query time (this is why it's disallowed by the spec).

See https://github.com/sourcegraph/sourcegraph/pull/19946 for a new validation pass to test this.

This PR combines both approaches. We'll only have one call to `indexDefinition` per symbol, but we don't need to write the range vertex if it already exists.

@slimsag 
Note that in future implementations that if you're creating any structure around a range that's used elsewhere (if you attach a result set, definition result, etc to it in the document symbols pass) you'll need to ensure that you don't create a second result set in `indexDefinition` (which assumes it's the first pass to construct that relationship).

### DIFF
--- a/internal/indexer/indexer.go
+++ b/internal/indexer/indexer.go
@@ -36,13 +36,14 @@ type Indexer struct {
 	vars    map[interface{}]*DefinitionInfo // position -> info
 
 	// LSIF data cache
-	documents             map[string]*DocumentInfo  // filename -> info
-	ranges                map[string]map[int]uint64 // filename -> offset -> rangeID
-	hoverResultCache      map[string]uint64         // cache key -> hoverResultID
-	packageInformationIDs map[string]uint64         // name -> packageInformationID
-	packageDataCache      *PackageDataCache         // hover text and moniker path cache
-	packages              []*packages.Package       // index target packages
-	projectID             uint64                    // project vertex identifier
+	documents             map[string]*DocumentInfo    // filename -> info
+	ranges                map[string]map[int]uint64   // filename -> offset -> rangeID
+	defined               map[string]map[int]struct{} // set of defined ranges (filename, offset)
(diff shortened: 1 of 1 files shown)


slimsag:
> Note that in future implementations that if you're creating any structure around a range that's used elsewhere (if you attach a result set, definition result, etc to it in the document symbols pass) you'll need to ensure that you don't create a second result set in indexDefinition (which assumes it's the first pass to construct that relationship).

Have I understood this correctly?

* For symbol and definition information to be attached to a given range, there must be a ResultSet associated with the range. There may be only one.
* `i.indexDefinition` creates a result set and then attaches the definition response to it.
* I will need to ensure symbol indexing reuses that result set - or that symbol indexing creates the result set for a range, attaches the symbol response to it, and then `indexDefinition` must attach its definition response to the same result set.

Is that right?

slimsag (approved):
Change makes sense, thanks for digging into this more.

efritz:
@slimsag Yes. If you end up creating _just_ a range vertex, that's fine. If you create a range vertex and document symbol vertexes and relationships, that's fine too. But if you create a resultSet and attach it to a range that will later be created by definition pass, you need to somehow cache those ids across indexing passes.

(And to be clear result sets are not a necessity but do make things easier to generate; we could attach everything directly to ranges, but that requires more space once serialized.)

slimsag:
Ahhh okay, that makes sense.

PR #145, opened 2021-05-03
Support replace directives in go.mod

efritz:
This PR adds support for `replace ()` directives inside of go.mod, as well as forks and canonical import paths in order to keep moniker identifiers stable. Note that because we're changing the format of monikers today, any index generated before this change will not be able to link across repositories to indexes generated after this change.

This changes the moniker scheme emitted by this indexer to emit a _canonical code host URL_ instead of the raw import path used by the dependent project. Raw import paths have several problems:

- Multiple import paths can resolve to the same repository (go.pkg.in and github.com can host the same content)
- The name declared by a go.mod file has no relation to the import path nor the repository it's contained within
- Multiple repositories can claim to provide the same module name (see Sourcegraph's fork of Zoekt)

Using a canonical code host URL (with a project suffix, if not at the repository root) solves these issues. We now use the same internal mechanism used by go get to resolve package name into something that can be cloned. The name of the module being indexed is determined from a mixture of a command line flag, the go.mod module declaration, and the name of the git remote (git remote is used as a base; the module name is used to determine a relative path within the repo).


internal/git/infer_repo.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+
+	"github.com/sourcegraph/lsif-go/internal/command"
+)
+
+// InferRepo gets a Sourcegraph-friendly repo name from the git clone enclosing
```
shrouxm, line 11: i think the comment here should say it's parsing it into lsif-go's expected format for packgeinformation rather than "sourcegraph-friendly"

garobrik (approved):
LGTM

PR #146, opened 2021-05-05
Do not emit duplicate import moniker vertices

efritz:
This change adds a cache of import moniker ids that are hashed by their identifier and package information relation. If a single imported identifier is used frequently in a dependent package (common, such as a logger or util functions) then we can use the same vertex and package information edge for each range/resultSet instead of re-emitting each time it's used.

This should reduce index size on large projects.

### DIFF
--- a/internal/indexer/moniker.go
+++ b/internal/indexer/moniker.go
@@ -17,19 +17,30 @@ func (i *Indexer) emitExportMoniker(sourceID uint64, p *packages.Package, obj ty
 		return
 	}
 
-	// If this is a package reference, remove the trailing `:`
-	monikerIdentifier := fmt.Sprintf("%s:%s", monikerPackage(obj), monikerIdentifier(i.packageDataCache, p, obj))
-	monikerIdentifier = strings.Trim(monikerIdentifier, ":")
+	// Emit export moniker (uncached as these are on unique definitions)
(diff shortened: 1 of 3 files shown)


internal/indexer/moniker.go
```diff
@@ -33,14 +40,20 @@ func (i *Indexer) emitImportMoniker(sourceID uint64, p *packages.Package, obj ty
...
-				"import",
-				strings.Trim(fmt.Sprintf("%s:%s", pkg, monikerIdentifier(i.packageDataCache, p, obj)), ":"),
-				sourceID,
-				i.ensurePackageInformation(moduleName, moduleVersion),
-			)
+		version, ok := i.dependencies[moduleName]
+		if ok {
+			// If this is a package reference, remove the trailing `:`
+			monikerIdentifier := fmt.Sprintf("%s:%s", pkg, monikerIdentifier(i.packageDataCache, p, obj))
+			monikerIdentifier = strings.Trim(monikerIdentifier, ":")
```
olafurpg, line 47: This trimming logic is duplicated in `emitExportMoniker` and `emitImportMoniker`. Might be worth consolidating since it's easy to forget to trim `:` resulting in hanging moniker IDs.

olafurpg (approved):
LGTM 👍

PR #148, opened 2021-05-06
Lookup cross-repo moniker path in correct package

efritz:
We were previously emitting monikers equivalent to the imported *package* rather than the moniker for the field within that package. This was because we were using the package of the _use_ rather than the _definition_ to look up the cached moniker path calculated in an earlier phase.

This lookup was used correctly where we extract hover text, meaning that jump to definition/find refs were broken while hover text was likely correct for a lot of situations. This makes the lookups symmetric.


garobrik (approved):
would be nice to have fewer one letter var names in lsif-go one day

LGTM! should we add a regression test for this?

PR #150, opened 2021-05-07
Implement Sourcegraph documentation extension 

slimsag:
This implements the Sourcegraph documentation extension to LSIF ([details on the extension here](https://github.com/sourcegraph/sourcegraph/pull/20108)

## What it can do

The implementation is already quite comprehensive and fast, you can get an idea of what it is capable of currently by looking at the `.golden` files in this PR:

* net_http.golden https://gist.github.com/slimsag/069d8498890ea6a3ec85dee3ef6f18c8
* regexp.golden https://gist.github.com/slimsag/c243dd016ab7afe0967450a61869ef97
* testdata.golden https://gist.github.com/slimsag/a8382ad1126e2d214659ed83f3a820f4

## Design choices

The implementation is based on walking the AST itself and pulling type information as needed. I chose to go this path instead of using e.g. the stdlib go/doc package (which is in maintenance mode and generally "deprecated" in the eyes of the Go developers in favor of pkg.go.dev, which is pretty WIP.) It also gives us a lot more flexibility in the future to do type correlation, etc. as well - which we will need in order to support hover, go-to-definition, and find-references on symbols within documentation type signatures in the future.

## Limitations

Hover/go-to-definition/find-references as outlined by the Sourcegraph documentation LSIF extension is not implemented here yet, my aim is to push a MVP of this through end-to-end and have it working on sourcegraph.com first before tackling that.

## Reviewing

There are two main things to look at:

* `internal/doctomarkdown` - a little helper package to convert LSIF documentation elements -> markdown format, for testing and debugging (since the raw JSON format is quite unconsumable as a human)
* `internal/indexer/documentation.go` - the actual implementation that emits the data

Fixes https://github.com/sourcegraph/sourcegraph/issues/19591
Helps https://github.com/sourcegraph/sourcegraph/issues/19389


internal/doctomarkdown/doctomarkdown.go
```diff
@@ -0,0 +1,296 @@
...
+	"io"
+	"strings"
+
+	"github.com/sourcegraph/sourcegraph/enterprise/lib/codeintel/lsif/protocol"
+	"github.com/sourcegraph/sourcegraph/enterprise/lib/codeintel/lsif/protocol/reader"
+)
+
+// Convert converts the LSIF data in r to Markdown by scanning for data in the Sourcegraph LSIF
+// documentation extension format.
+func Convert(ctx context.Context, r io.Reader, matchingTags []protocol.DocumentationTag) (string, error) {
```
efritz, line 18: It looks like this entire package is used only for testing. Do you plan to make this available more generally? If not, I'd probably make this an internal test file in the indexer package. If you do make it more available generally (sharing code with the internal conversion process), then that code should be public and in sg/sg/lib.

internal/indexer/documentation.go
```diff
@@ -0,0 +1,748 @@
...
+		d            = &docsIndexer{i: i}
+		docsPackages []docsPackage
+		errs         error
+	)
+	i.visitEachPackage("Indexing documentation", func(p *packages.Package) {
+		docsPkg, err := d.indexPackage(p)
+		if err != nil {
+			errs = multierror.Append(errs, errors.Wrap(err, "package "+p.Name))
+		}
+		docsPackages = append(docsPackages, docsPkg)
```
efritz, line 47: You'll throw on line 65 later when `err != nil`

internal/indexer/documentation.go
```diff
@@ -0,0 +1,748 @@
...
+		return v.String()
+	}
+}
+
+// isDeprecated tells if the given docstring has any line beginning with "deprecated",
+// "Deprecated", or "DEPRECATED".
+func isDeprecated(docstring string) bool {
+	for _, line := range strings.Split(docstring, "\n") {
+		for _, deprecated := range []string{
+			"deprecated",
```
efritz, line 685: Why not just fold case before the comparison?

internal/indexer/documentation.go
```diff
@@ -0,0 +1,748 @@
...
+	err := format.Node(&buf, fset, node)
+	if err != nil {
+		panic(fmt.Sprintf("never here: %v", err))
+	}
+	return buf.String()
+}
+
+// formatTypeLabel formats the given type as a one-line label, omitting certain details like struct
+// fields.
+func formatTypeLabel(t types.Type) string {
```
efritz, line 633: Can we get specific unit tests for these util functions?

internal/indexer/documentation.go
```diff
@@ -0,0 +1,748 @@
...
+	// Produce the full type signature with docs on e.g. methods and struct fields, but not on the
+	// type itself (we'll produce those as Markdown below.)
+	cpy := *in
+	cpy.Doc = nil
+	result.signature = typ + " " + formatNode(p.Fset, &cpy)
+
+	// TODO(slimsag): future: this is a HACK because some variables/constants are ultra long table
+	// initializers, including those is not helpful to the user so we fallback in this case to
+	// something much briefer.
+	if (strings.HasSuffix(result.signature, "}\n") || strings.HasSuffix(result.signature, "}")) && strings.Contains(result.signature, "{") {
```
efritz, line 415: Could we just check the length instead?

internal/indexer/indexer.go
```diff
@@ -108,6 +108,7 @@ func (i *Indexer) Index() error {
 	i.emitMetadataAndProjectVertex()
 	i.emitDocuments()
 	i.addImports()
+	i.indexDocumentation()
```
efritz, line 111: Could we enable this via flag since it's currently a non-published extension?

efritz (approved):
I don't have any major concerns here, just some notes inline.

internal/indexer/documentation.go
```diff
@@ -0,0 +1,748 @@
...
+	if f.deprecated {
+		tags = append(tags, protocol.DocumentationDeprecated)
+	}
+
+	// Include the full type signature
+	var detail bytes.Buffer
+	fmt.Fprintf(&detail, "```Go\n")
+	fmt.Fprintf(&detail, "%s\n", f.signature)
+	fmt.Fprintf(&detail, "```\n\n")
+	fmt.Fprintf(&detail, "%s", f.docsMarkdown)
```
Strum355, line 534:
```suggestion
	var detail strings.Builder
	detail.WriteString("```Go\n")
	detail.WriteString("%s\n", f.signature)
	detail.WriteString("```\n\n")
	detail.WriteString("%s", f.docsMarkdown)
```

using strings.Builder would most likely be more efficient both from memory and cpu perspective here than a bytes.Buffer through fmt.Fprintf; calls to the bytes.Buffer method are probably virtualized as fmt.Fprintf takes an io.Writer interface, and the strings passed might be copied as they are converted to []byte, but I havent confirmed that last part. With strings.Builder, there is some magic to avoid copying of the string.

Even better I believe would be basic string concatenation, as these are straight combining (as opposed to more complex control flow) so the compiler can most likely make this an optimal operation. I will write some benchmarks out of curiosity :slightly_smiling_face:

Strum355: Without `b.Grow(...)`, string concat outperforms bytes.Buffer and strings.Builder, but with it, strings.Builder comes out on top 
image

Strum355:
```suggestion
	var detail strings.Builder
	detail.Grow(6 + len(f.signature) + len(f.docsMarkdown) + 6)
	detail.WriteString("```Go\n")
	detail.WriteString(f.signature)
	detail.WriteRune('\n')
	detail.WriteString("```\n\n")
	detail.WriteString(f.docsMarkdown)
```

internal/doctomarkdown/doctomarkdown.go
slimsag: just for testing, will move.

internal/indexer/indexer.go
slimsag: what needs to be done to un-flag it in your view? doesn't it being behind a flag mean off-by-default?

internal/indexer/documentation.go
slimsag: I'm going to revisit this later, I don't think spending more time making this specific check smarter (or less smart) makes sense at this point. Rather just fix it for real. :)

internal/indexer/indexer.go
efritz: Off-by-default makes sense today since we don't have a backend that can consume it, so we'd be inflating index time for customers that can't make use of the new data yet. I'd say make it on-by-default (or always on) once we have a version that ships the feature.

For auto-indexing on cloud and for our own code we can put the flag explicitly for the next month or so.

slimsag: My aim is to only merge this PR once I have the actual Sourcegraph backend consuming this data and optionally rendering it behind a feature flag in the site config. Given that, do you still think it should be off-by-default here too (two feature flags, effectively)?

> For auto-indexing on cloud and for our own code we can put the flag explicitly for the next month or so.

Where would this be done? I'm not familiar with e.g. where auto indexing gets invoked.

efritz: Actually if we plan to consume it before we merge this I'm fine just emitting it always.

internal/indexer/documentation.go
slimsag: Right now this does have quite good test coverage through the `child_symbols.go` + markdown regression/integration tests, but I agree it'd be nice to have unit tests for these util functions as well. I am going to follow up on this and am tracking this here: #20567

internal/indexer/documentation.go
slimsag: Thank you!

PR #151, opened 2021-05-10
Move from MarkedString to MarkupContent 

slimsag:
This should successfully update enterprise/lib to unblock me on #150 and update lsif-go to use the new MarkupContent API.

* sourcegraph/sourcegraph#20284 changed the enterprise/lib package to _only_ support emitting MarkupContent, but lsif-go depended on being able to emit MarkedString (the deprecated API.)
* Noah had a WIP commit to update lsif-go to emit MarkupContent instead of MarkedString, but it had some test failures and the upstream protocol writer package did not have support for writing MarkupCOntent.
* In sourcegraph/sourcegraph#20830 I added support for _writing_ MarkupContent.


internal/indexer/protocol.go
```diff
@@ -30,39 +30,40 @@ func rangeForObject(obj types.Object, pos token.Position) (protocol.Pos, protoco
...
 		}
 	}
 
-	return mss
+	return protocol.NewMarkupContent(strings.Join(ss, "\n\n---\n\n"), protocol.Markdown)
 }
 
-// formatMarkdown creates a protocol.MarkedString object containing a markdown-formatted version
+// formatMarkdown creates a string containing a markdown-formatted version
 // of the given string. If the given string is empty, nil is returned.
```
Strum355, line 49: just as a nit, here and L61 weren't updated (by me :hand:) to reflect the change in what gets returned

Strum355 (approved):

PR #153, opened 2021-05-18
Improve go.mod-less indexing experience

efritz:
Fix name of exported package in projects without a go.mod file. With `GO111MODULE=off`, the packages being indexed have names like `_/path/to/this/project` instead of the usual import path. In the case of no go.mod file, we use a pruned name of the remote repository which we infer to be a valid import path. We use this value instead when the package begins with `_{indexing path}`.

### DIFF
--- a/internal/indexer/moniker_test.go
+++ b/internal/indexer/moniker_test.go
@@ -15,7 +15,10 @@ func TestEmitExportMoniker(t *testing.T) {
 	w := &capturingWriter{}
 
 	indexer := &Indexer{
-		moduleName:            "github.com/sourcegraph/lsif-go",
+		repositoryRemote:      "github.com/sourcegraph/lsif-go",
+		repositoryRoot:        "/users/efritz/dev/sourcegraph/lsif-go",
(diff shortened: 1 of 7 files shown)


internal/indexer/moniker_test.go
```diff
@@ -15,7 +15,10 @@ func TestEmitExportMoniker(t *testing.T) {
 	w := &capturingWriter{}
 
 	indexer := &Indexer{
-		moduleName:            "github.com/sourcegraph/lsif-go",
+		repositoryRemote:      "github.com/sourcegraph/lsif-go",
+		repositoryRoot:        "/users/efritz/dev/sourcegraph/lsif-go",
+		projectRoot:           "/users/efritz/dev/sourcegraph/lsif-go",
+		moduleName:            "https://github.com/sourcegraph/lsif-go",
```
shrouxm, line 21: how come the module name becomes `https://...`?

garobrik (approved):
LGTM

internal/indexer/moniker_test.go
efritz: I forgot to update the test when we change the moniker schemes for v1.4.0.

PR #163, opened 2021-06-15
API docs: update to latest protocol spec

slimsag:
This is a few changes (sent as one because they all depend on each-other, sorry):

* Updating `sourcegraph/enterprise/lib` and dealing with the fact it moved to `sourcegraph/lib`
* Updating the API docs indexing code to emit according to the revised spec: https://github.com/sourcegraph/sourcegraph/pull/22080


efritz (approved):
Thanks for 500 lines being go.sum and the golden test 😃

PR #165, opened 2021-06-16
 awkward: some of our testdata Go files have syntax errors!

slimsag:
```
$ find internal/testdata -type f -name '*.go' -exec go fmt {} \;
internal/testdata/child_symbols.go:67:2: expected declaration, found '}'
exit status 2
```

I didn't realize this because the Go parser is surprisingly good (and silent)
about handling file parsing errors, and `go fmt ./...` ignores `testdata/` files.

This adds a CI check which ensures our files are syntactically valid at least.

### DIFF
--- a/buildkite.yml
+++ b/buildkite.yml
@@ -1,3 +1,5 @@
 steps:
   - label: ':go:'
+    command: find internal/testdata -type f -name '*.go' -exec go fmt {} \;
+  - label: ':go:'
     command: go test ./...
--- a/internal/testdata/child_symbols.go
+++ b/internal/testdata/child_symbols.go
@@ -42,7 +42,7 @@ var BigVar Interface = &Struct{
 //
 // It's sleeping! Some people write that as `sleeping` but Markdown
 // isn't allowed in Go docstrings, right? right?!
-var(
+var (
 	// This has some docs
 	VarBlock1 = "if you're reading this"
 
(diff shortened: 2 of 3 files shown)


buildkite.yml
```diff
@@ -1,3 +1,5 @@
 steps:
+  - label: ':go:'
+    command: find internal/testdata -type f -name '*.go' -exec go fmt {} \;
```
efritz, line 3: Would go fmt ./... be as sufficient?

internal/testdata/child_symbols.go
```diff
@@ -89,9 +89,9 @@ func (s *Struct) MachineLearning(
 	// 	   `--'   `--'
 	//
 	hyperparam2 float32,
-	hyperparam3 float32
+	hyperparam3 float32,
 ) float32 {
-	return param1+(hyperparam2**hyperparam3) // lol is this all ML is? I'm gonna be rich
+	return param1 + (hyperparam2 * *hyperparam3) // lol is this all ML is? I'm gonna be rich
```
efritz, line 94: lofl

buildkite.yml
slimsag: Nope, `go fmt ./...` excludes testdata directories! :(

efritz: TIL

efritz (approved):

buildkite.yml
efritz: Just to clarify, is the return value of this non-zero whenever there's a change on disk?

slimsag: oops, good catch - it's not. Will fix.

PR #166, opened 2021-06-16
API docs: only index top-level symbols

slimsag:
This adds a failing test showing that non-top-level symbols are currently indexed by API docs, and then fixes the issue. Using an AST visitor was definitely incorrect here, not sure how I missed it :)

Fixes sourcegraph/sourcegraph#22030


buildkite.yml
```diff
@@ -1,3 +1,5 @@
 steps:
+  - label: ':go:'
+    command: find internal/testdata -type f -name '*.go' -exec go fmt {} \;
```
efritz, line 3: Stacked PR?

slimsag: yeah, rebasing now.

internal/indexer/documentation.go
```diff
@@ -299,13 +299,9 @@ func (d *docsIndexer) indexFile(p *packages.Package, f *ast.File, isTestFile boo
...
 	result.pkgDocsMarkdown = godocToMarkdown(f.Doc.Text())
 
-	// Visit each AST node, collecting the bits we want to emit documentation for.
-	var (
-		firstErr error
-		visitor  ast.Visitor
-	)
-	visitor = visitorFunc(func(n ast.Node) (w ast.Visitor) {
-		switch node := n.(type) {
+	// Collect each top-level declaration.
```
efritz, line 302: Oh this is probably much faster as well than walking the entire AST.

efritz (approved):

internal/indexer/documentation.go
slimsag: yeah :D

PR #167, opened 2021-06-17
API docs: indexer: do a better job of trimming long const/var assignments

slimsag:


internal/indexer/documentation.go
```diff
@@ -454,7 +454,7 @@ func (d *docsIndexer) indexConstVar(p *packages.Package, in *ast.ValueSpec, name
 	// TODO(slimsag): future: this is a HACK because some variables/constants are ultra long table
 	// initializers, including those is not helpful to the user so we fallback in this case to
 	// something much briefer.
-	if (strings.HasSuffix(result.signature, "}\n") || strings.HasSuffix(result.signature, "}")) && strings.Contains(result.signature, "{") {
+	if len(result.signature) > 100 {
```
slimsag, line 457: Turns out your original suggestion was indeed a lot better :)

efritz (approved):

PR #184, opened 2021-07-21
wip: add stdlib dependencies

tjdevries:
Just adding this so I don't accidentally delete some stuff that I wrote... :'(


internal/gomod/dependencies.go
```diff
@@ -61,17 +63,24 @@ type jsonModule struct {
 	Name    string      `json:"Path"`
 	Version string      `json:"Version"`
 	Replace *jsonModule `json:"Replace"`
+
+	// The Golang version required for this module
+	GoVersion string `json:"GoVersion"`
 }
 
+var golangRepository = "github.com/golang/go"
+var validGolangVersion = regexp.MustCompile(`(\d+)\.(\d+)`)
```
efritz, line 72: `<(^^<) (>^^)>`

internal/gomod/dependencies.go
```diff
@@ -95,15 +120,88 @@ func parseGoListOutput(output, rootVersion string) (map[string]Module, error) {
...
+	setGolangDependency(dependencies, goVersion)
+
 	return dependencies, nil
 }
 
+type parsedGoVersion struct {
+	Version string
+	Major   int
+	Minor   int
+}
```
efritz, line 141: We use a semver package in sg/sg that might be useful here as well. Saves you from parsing the version manually here.

internal/gomod/dependencies.go
```diff
@@ -95,15 +120,88 @@ func parseGoListOutput(output, rootVersion string) (map[string]Module, error) {
...
+		Version: version,
+		Major:   major,
+		Minor:   minor,
+	}, nil
+}
+
+func setGolangDependency(dependencies map[string]GoModule, goVersion parsedGoVersion) {
+	dependencies[golangRepository] = GoModule{
+		Name:    golangRepository,
+		Version: "go" + goVersion.Version,
```
efritz, line 166: Can we add a comment stating this is tied to golang/go's git tag scheme?

internal/gomod/dependencies.go
```diff
@@ -95,15 +120,88 @@ func parseGoListOutput(output, rootVersion string) (map[string]Module, error) {
...
+}
+
+// IsStandardlibPackge checks whether a particular package is a standard
+// library package.
+func IsStandardlibPackge(pkg string) bool {
+	if strings.Contains(pkg, ".") {
+		return false
+	}
+
+	if strings.HasPrefix(pkg, "_") {
```
efritz, line 182: When does this happen?

internal/gomod/dependencies.go
```diff
@@ -95,15 +120,88 @@ func parseGoListOutput(output, rootVersion string) (map[string]Module, error) {
...
+}
+
+func GetGolangDependency(dependencies map[string]GoModule) GoModule {
+	return dependencies[golangRepository]
+}
+
+// IsStandardlibPackge checks whether a particular package is a standard
+// library package.
+func IsStandardlibPackge(pkg string) bool {
+	if strings.Contains(pkg, ".") {
```
efritz, line 178: Can we describe the heuristics here? "." is because it lacks a domain name in the import path.

internal/indexer/moniker.go
```diff
@@ -136,17 +137,20 @@ func (i *Indexer) ensureImportMoniker(identifier string, packageInformationID ui
...
+	} else {
+		pkgName = obj.Pkg().Path()
 	}
 
-	return obj.Pkg().Path()
+	return gomod.NormalizeMonikerPackage(pkgName)
 }
 
-// monikerIdentifier returns the identifier suffix used to construct a unique moniker for the given object.
+// getMonikerIdentifier returns the identifier suffix used to construct a unique moniker for the given object.
```
efritz, line 150: Why change this (but not `monikerPackage`)?

internal/gomod/dependencies.go
```diff
@@ -95,15 +120,88 @@ func parseGoListOutput(output, rootVersion string) (map[string]Module, error) {
...
 
-		dependencies[importPath] = Module{
+		dependencies[importPath] = GoModule{
 			Name:    module.Name,
 			Version: cleanVersion(module.Version),
 		}
 	}
 
+	if goVersion.Version == "" {
+		return nil, errors.New("Must have a valid go version in go mod file")
```
efritz, line 130: Errors must start with lower-case letters (and not end in punctuation).

internal/gomod/dependencies.go
tjdevries: <discuessed in slack. Useful for testing purposes. Also, no std lib packages start with "_">

internal/gomod/dependencies.go
```diff
@@ -95,15 +107,76 @@ func parseGoListOutput(output, rootVersion string) (map[string]Module, error) {
...
+		return false
+	}
+
+	// No standard library packages are prefixed with "_".
+	// This can be used to denote testing packages.
+	if strings.HasPrefix(pkg, "_") {
+		return false
+	}
+
+	return true
```
Strum355, line 161: Just to really ruin this for ya, its entirely possible and legit for non-stdlib fully-qualified import path to _not_ be a domain name. Youd usually only see this in application repos, as you cant `go get` them for use as libraries

Strum355: Conferring with a Go compiler dev, a possibility would be to invoke `go list std` and seeing if the import name is in that list or not

tjdevries: oh, `go list std` is way easier. Only thing would be multiple different versions, but we can update that manually over time or something.

tjdevries:
OK, I implemented a possible version that we could do. In like... 15 more minutes or so I can probably add a bunch of the other go versions. If we really wanted, I could write something that generates files and we can use that magic hack of `go:embed` and then run some docker thing that will generate the files from different go versions.

Anyway, I can do whatever :)

Strum355:
Given we already invoke the Go cli for other purposes, it might be easier and less maintenance burden to invoke it for `go list std`? 🙂

tjdevries:
If your go program says its version is 1.14 and you are using the go 1.16, that would be very confusing to get different lsif dumps IMO. I don't like that it can be non-deterministic based on the compiler version you ran it with.

Strum355:
Important to note is that the `go` directive in go.mod files only indicates the Go version when it comes to _language_ features. It doesnt restrict using stdlib packages that were introduced at a later stage.

e.g. `io.ReadAll` was added in 1.16, but its completely legit to have a go.mod that references 1.14 as the minimum version and use `io.ReadAll`, assuming the environment has a 1.16+ compiler.
What it would disallow is using underscore delimited numbers like 1_000_000, introduced in 1.13, when the `go` directive states the language features

internal/indexer/moniker.go
efritz: It's not super idiomatic in Go to have `getX`, `setX` methods. My biggest concern is consistency, though.

tjdevries:
But if someone had `io.ReadAll` in their code, it wouldn't compile with a 1.14 compiler, right?

Strum355:
Correct, but its completely legit to have your go.mod reference "1.14", while the environment has a 1.16 compiler and therefore use `io.ReadAll`.

tjdevries (commented):
OK, fixed up what we were talking about for the go list stuff.

Basically we settled on whatever the latest version of `go` currently is, we'll run `go list std` from that command and then store that as a fixed list so that we don't have any non-deterministic output for which import strings are stdlib or not.

Additionally, I changed things to just run one more command to get the current packages `GoVersion` information, so that we can just pull whatever the go.mod file says the package information is. I don't understand why Go is the way it is in this way, but let's just let people update their own `go.mod` files to the version they expect to run this with.

internal/gomod/dependencies.go
```diff
@@ -95,15 +105,59 @@ func parseGoListOutput(output, rootVersion string) (map[string]Module, error) {
...
+}
+
+func GetGolangDependency(dependencies map[string]GoModule) GoModule {
+	return dependencies[golangRepository]
+}
+
+// NormalizeMonikerPackage returns a normalized path to ensure that all
+// standard library paths are handled the same. Primarily to make sure
+// that both the golangRepository and "std/" paths are normalized.
+func NormalizeMonikerPackage(path string) string {
```
efritz, line 148: Can we get unit tests for this new function?

internal/gomod/stdlib.go
```diff
@@ -0,0 +1,232 @@
...
+
+// This list is calculated from "go list std".
+//
+// You can get a formattable golang snippet by:
+//
+//    $ echo "var standardLibraryMap = map[string]interface{}{" \
+//       && go list std | awk '{ print "\""$0"\": contained,"}' \
+//       && echo "}"
+//
+var standardLibraryMap = map[string]interface{}{
```
efritz, line 21: Can we define this in a file on its own, and add a bash script to the repo that dumps that content to the file?

internal/gomod/stdlib.go
```diff
@@ -0,0 +1,232 @@
...
+
+// IsStandardlibPackge determines whether a package is in the standard library
+// or not. At this point, it checks whether the package name is one of those
+// that is found from running "go list std" in the latest released go version.
+func IsStandardlibPackge(pkg string) bool {
+	_, ok := standardLibraryMap[pkg]
+	return ok
+}
+
+var contained interface{}
```
efritz, line 11: `var contained struct{}{}` is a bit more idiomatic (and you can inline it since it's a zero-width type).

internal/gomod/stdlib.go
efritz: Also, nice strategy. Thanks @tjdevries and @Strum355 for working through these tricky edge cases!

tjdevries: ty ty :) added script to generate completely

internal/gomod/dependencies.go
tjdevries: yup! done

efritz (approved):
🎉 Works so clean so fresh

PR #186, opened 2021-08-10
Switch to alpine as base image for lsif-go

eseliger:
This reduces the image size from 841MB to 346MB, should give a significant boost in startup time.
Q: Do we rely on any tools other than git to be installed?


Strum355 (approved):
LGTM, deferring to eric on your Q

efritz (commented):
We only shell out to git and go in lsif-go. Could this be problematic for builds that require CGO?

chrispine:
Is this still something being worked on?

eseliger:
Not sure, I'll close it for now.

PR #188, opened 2021-08-11
feat: Goto definition of package and imports

tjdevries:


internal/indexer/typestring.go
```diff
@@ -26,10 +26,19 @@ func typeString(obj types.Object) (signature string, extra string) {
...
+
+	}
+
+	// Fall back to types.Object if possible
+	if v, ok := obj.(types.Object); ok {
+		return types.ObjectString(v, packageQualifier), ""
 	}
 
-	return types.ObjectString(obj, packageQualifier), ""
+	return "TODO DONT LET TJ MERGE THIS", ""
```
efritz, line 41: - [ ] Don't let TJ merge this

internal/testdata/child_symbols.go
```diff
@@ -1,4 +1,3 @@
-// Package testdata
```
efritz, line 1: Is this removed on purpose?

internal/indexer/testdata/TestIndexer_documentation/testdata/cmd.json
```diff
@@ -0,0 +1,22 @@
+{
```
efritz, line 1: What is this file?

internal/indexer/moniker.go
```diff
@@ -136,10 +136,11 @@ func (i *Indexer) ensureImportMoniker(identifier string, packageInformationID ui
 
 // makeMonikerPackage returns the package prefix used to construct a unique moniker for the given object.
 // A full moniker has the form `{package prefix}:{identifier suffix}`.
-func makeMonikerPackage(obj types.Object) string {
+func makeMonikerPackage(obj ObjectLike) string {
 	var pkgName string
 	if v, ok := obj.(*types.PkgName); ok {
-		pkgName = strings.Trim(v.Name(), `"`)
+		// pkgName = strings.Trim(v.Name(), `"`)
+		pkgName = v.Imported().Path()
```
efritz, line 143: Can I get an example of what this changes? (Also, delete the old commented code if you're confident in this new expression).

internal/testdata/child_symbols.go
tjdevries: Yes, there should only be one spot that has a comment for the package `testdata` otherwise it's weird for choosing the documentation hover string for the package.

internal/indexer/testdata/TestIndexer_documentation/testdata/cmd.json
tjdevries: No idea, it got autogenerated by the tests, so I thought it was supposed to be committed haha

internal/indexer/package_declaration.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"go/token"
+	"go/types"
+
+	"golang.org/x/tools/go/packages"
+)
+
+// TODO: Handle testing packages differently.
+
+type ObjectLike interface {
```
efritz, line 12: Why was it necessary to define this over `types.Object`?

internal/indexer/indexer.go
```diff
@@ -280,36 +291,78 @@ func (i *Indexer) emitDocument(filename string) {
...
+//              ^^^---- reference github.com/golang/go/std/fmt
+//
+// In both cases, this will emit the corresponding import moniker for "fmt". This is ImportSpec.Path
+func (i *Indexer) emitImportMonikerReference(p *packages.Package, pkg *packages.Package, spec *ast.ImportSpec) {
+	state := makePkgNameState(i, p, spec.Path.Pos(), spec.Path.Value, pkg)
+
+	i.emitImportMoniker(state.resultSetID, p, state.obj)
+
+	// @eric -- we could potentially skip this because we have am oniker now.
+	// but this breaks a test and the current structure.
```
efritz, line 316: I think it's fine to emit the text here - we cache the hover results and we already emit it on any package reference so worrying about it _here_ doesn't fix much.

We could try to drop all cached hover references to external packages assuming we have good enough index coverage in the wild in the future.

internal/indexer/indexer.go
```diff
@@ -456,16 +518,20 @@ func (i *Indexer) indexDefinition(p *packages.Package, filename string, document
...
 
 	if obj.Exported() {
 		i.emitExportMoniker(resultSetID, p, obj)
 	}
 
 	// If the pkg/object has associated documentation, link to it. This enables e.g. going from documentation
 	// for a symbol <-> its definition/hover/references/etc in either direction.
+	//
+	// @eric I'm not sure we need to even do this anymore.
+	// We have the monikers correct now, so I don't think there is any reason to even do anything for these (maybe?)
```
efritz, line 534: Let's keep as much in-index as we can for now until we're sure we have sufficient index coverage and can rely on x-repo lookups of documentation/hover text.

Currently our schema is better suited for in-index queries, as the other path has additional database interactions.

internal/indexer/indexer.go
```diff
@@ -456,16 +518,20 @@ func (i *Indexer) indexDefinition(p *packages.Package, filename string, document
 		}))
 	}
 
-	if _, ok := obj.(*types.PkgName); ok {
-		i.emitImportMoniker(resultSetID, p, obj)
-	}
+	// @eric, I don't think we want this to happen anymore
+	// if _, ok := obj.(*types.PkgName); ok {
+	// 	i.emitImportMoniker(resultSetID, p, obj)
+	// }
```
efritz, line 524: 👍  Seems legit

internal/indexer/indexer.go
```diff
@@ -612,7 +700,7 @@ func (i *Indexer) indexReferenceToDefinition(p *packages.Package, document *Docu
 	d.m.Unlock()
 
 	if d.TypeSwitchHeader {
-		// Attache a hover text result _directly_ to the given range so that it "overwrites" the
```
efritz, line 615: image

internal/indexer/indexer.go
```diff
@@ -642,21 +730,23 @@ func (i *Indexer) indexReferenceToExternalDefinition(p *packages.Package, docume
...
-	_ = i.emitter.EmitTextDocumentReferences(rangeID, refResultID)
-	_ = i.emitter.EmitItemOfReferences(refResultID, []uint64{rangeID}, document.DocumentID)
-
 	if hoverResultID != 0 {
 		_ = i.emitter.EmitTextDocumentHover(rangeID, hoverResultID)
 	}
 
+	// You might think, "Hey, TJ, don't you want to emit a textDocument/references result here?"
+	//   Great question!
+	//   No.
```
efritz, line 739: Why not?

internal/indexer/indexer.go
```diff
@@ -723,6 +813,115 @@ func (i *Indexer) emitContainsForProject() {
...
+}
+
+func newPkgDeclaration(p *packages.Package, f *ast.File) (*PkgDeclaration, token.Position) {
+	pkgKeywordPosition := p.Fset.Position(f.Package)
+
+	name := f.Name.Name
+
+	// TODO(packages): Is this the right package information here? I'm not sure.
+	// obj := types.NewPkgName(f.Package, p.Types, name, types.NewPackage(p.PkgPath, name))
+	return &PkgDeclaration{
```
efritz, line 827: Do you have to construct a type here? Can't you get it from `f.Syntax` somewhere?

internal/indexer/indexer.go
```diff
@@ -723,6 +813,115 @@ func (i *Indexer) emitContainsForProject() {
...
+
+		// Skip everything that isn't the best
+		if position.Filename != best {
+			continue
+		}
+
+		name := obj.Name()
+		_, d, ok := i.positionAndDocument(p, obj.Pos())
+		if !ok {
+			panic("hmmm, probably should not merge this code")
```
efritz, line 881: - [ ] lol

efritz: I'd log a warning and continue - ask the warning to file a bug?

internal/indexer/indexer.go
```diff
@@ -732,3 +931,69 @@ func (i *Indexer) Stats() IndexerStats {
...
+
+func (i *Indexer) setRangeForPosition(position token.Position, id uint64) {
+	i.stripedMutex.LockKey(position.Filename)
+	i.ranges[position.Filename][position.Offset] = id
+	i.stripedMutex.UnlockKey(position.Filename)
+}
+
+// TODO:
+// - We should sort by edit distance here, so that it picks the best name we can as a last resort.
+func findBestPackageDefinitionPath(packageName string, possiblePaths []DeclInfo) (string, error) {
```
efritz, line 943: 🤤

internal/indexer/moniker.go
tjdevries: Yup, added comment explaining.

internal/indexer/package_declaration.go
tjdevries: We have to make our own type, PkgDeclaration. I will add explanation above the struct

internal/indexer/indexer.go
tjdevries: If we make a reference here, we won't look up the import moniker, which is what we really want. That will prevent linking these correctly between packages.

internal/indexer/indexer.go
tjdevries: It has no f.Syntax type, it is a conglomeration of fields from `f.` that tells us the information. For example, the position is just stored in `f.Package`.

internal/indexer/indexer.go
efritz: ok - can you put this in the comment? 🙂

internal/indexer/indexer.go
```diff
@@ -345,14 +398,16 @@ func (i *Indexer) indexDefinitionsForPackage(p *packages.Package) {
...
+	caseClauses := map[token.Pos]ObjectLike{}
 	for node, obj := range p.TypesInfo.Implicits {
 		if _, ok := node.(*ast.CaseClause); ok {
 			caseClauses[obj.Pos()] = obj
 		}
 	}
 
-	for ident, obj := range p.TypesInfo.Defs {
+	for ident, typeObj := range p.TypesInfo.Defs {
+		var obj ObjectLike = typeObj
```
efritz, line 409: Why is this cast necessary?

internal/indexer/indexer.go
```diff
@@ -732,3 +937,93 @@ func (i *Indexer) Stats() IndexerStats {
...
+
+func fileNameWithoutExtension(fileName string) string {
+	return strings.TrimSuffix(fileName, path.Ext(fileName))
+}
+
+func filterBasedOnTestFiles(possiblePaths []DeclInfo, packageName string) []DeclInfo {
+	preferredPaths := []DeclInfo{}
+	if !strings.HasSuffix(packageName, "_test") {
+		for _, v := range possiblePaths {
+			if !strings.HasSuffix(v.Path, "_test.go") {
```
efritz, line 1012: You could collapse these loops with the condition:

`if strings.HasSuffix(packageName, "_test") == strings.HasSuffix(v.Path, "_test.go")`

internal/indexer/indexer.go
```diff
@@ -723,6 +845,89 @@ func (i *Indexer) emitContainsForProject() {
...
+
+type DeclInfo struct {
+	HasDoc bool
+	Path   string
+}
+
+func (i *Indexer) indexPackageDeclarationForPackage(p *packages.Package) {
+	// Pick the filename that is the most idiomatic for the defintion of the package.
+	// This will make jump to def always send you to a better go file than the $PKG_test.go, for example.
+	packageDeclarations := []DeclInfo{}
```
efritz, line 860:
```suggestion
	packageDeclarations := make([]DeclInfo, 0, len(p.Syntax))
```

internal/indexer/indexer.go
```diff
@@ -723,6 +845,89 @@ func (i *Indexer) emitContainsForProject() {
...
+	i.visitEachPackage("Indexing package declarations", i.indexPackageDeclarationForPackage)
+}
+
+type DeclInfo struct {
+	HasDoc bool
+	Path   string
+}
+
+func (i *Indexer) indexPackageDeclarationForPackage(p *packages.Package) {
+	// Pick the filename that is the most idiomatic for the defintion of the package.
```
efritz, line 858: Can this comment be move down to directly above the call?

internal/indexer/indexer.go
```diff
@@ -723,6 +845,89 @@ func (i *Indexer) emitContainsForProject() {
...
+func (i *Indexer) indexPackageDeclarations() {
+	i.visitEachPackage("Indexing package declarations", i.indexPackageDeclarationForPackage)
+}
+
+type DeclInfo struct {
+	HasDoc bool
+	Path   string
+}
+
+func (i *Indexer) indexPackageDeclarationForPackage(p *packages.Package) {
```
efritz, line 857: Thank you for the inline comments in this function. They were very thoughtful and were helpful to navigate when unfamiliar with the code.

internal/indexer/indexer.go
```diff
@@ -642,21 +764,21 @@ func (i *Indexer) indexReferenceToExternalDefinition(p *packages.Package, docume
 	})
 
 	rangeID, _ := i.ensureRangeFor(pos, definitionObj)
-	refResultID := i.emitter.EmitReferenceResult()
-	_ = i.emitter.EmitTextDocumentReferences(rangeID, refResultID)
-	_ = i.emitter.EmitItemOfReferences(refResultID, []uint64{rangeID}, document.DocumentID)
-
```
efritz, line 648: @tjdevries was putting this back the solution to the one-way navigation issue we were seeing?

internal/indexer/indexer.go
```diff
@@ -149,10 +156,17 @@ func (i *Indexer) loadPackages(deduplicate bool) error {
 			Logf:  i.packagesLoadLogger,
 		}
 
-		pkgs, err := packages.Load(config, "./...")
-		if err != nil {
-			errs <- errors.Wrap(err, "packages.Load")
-			return
+		// Make sure we only load packages once per execution.
+		pkgs, ok := cachedPackages[i.projectRoot]
```
efritz, line 160: @tjdevries is this was what you referring to as future-proofing other calls to `packages.Load`?

tjdevries: Yeah, that's the one.

internal/indexer/indexer.go
tjdevries: I fixed this by doing the linking now

tjdevries: ( linkImportMonikersToRanges )

internal/indexer/moniker.go
```diff
@@ -53,7 +53,8 @@ func joinMonikerParts(parts ...string) string {
 // identifier (either a range or a result set identifier). This will also emit links between
 // the moniker vertex and the package information vertex representing the dependency containing
 // the identifier.
-func (i *Indexer) emitImportMoniker(sourceID uint64, p *packages.Package, obj types.Object) {
+// TODO THINGS
```
efritz, line 56: What things?

internal/indexer/indexer.go
```diff
@@ -127,8 +148,54 @@ func (i *Indexer) Index() error {
 	return nil
 }
 
+func (i *Indexer) startImportMonikerReferenceTracker(wg *sync.WaitGroup) {
+	go func() {
+		contained := struct{}{}
+
+		wg.Add(1)
```
efritz, line 155: `wg.Add(1)` must be called _outside_ of the goroutine that you are protecting, otherwise the scheduler can see a zero count before the goroutine is scheduled and exit.

internal/indexer/indexer.go
```diff
@@ -127,8 +148,54 @@ func (i *Indexer) Index() error {
...
+		for {
+			nextReference := <-i.importMonikerChannel
```
efritz, line 157: You should be able to do `for nextReference := range i.ilmportMonikerChannel`, which reads until the channel is closed. This will get rid of the `done` field in your reference struct and simplify the condition below as well.

internal/indexer/indexer.go
```diff
@@ -696,6 +893,35 @@ func (i *Indexer) linkItemsToDefinitions(d *DefinitionInfo) {
 	}
 }
 
+func (i *Indexer) linkImportMonikersToRanges(wg *sync.WaitGroup) {
+	i.stopImportMonikerReferenceTracker(wg)
```
efritz, line 897: I'd rather have this in the main indexer method - this flow was non-intuitive to me when reading it.

internal/indexer/indexer.go
```diff
@@ -696,6 +893,35 @@ func (i *Indexer) linkItemsToDefinitions(d *DefinitionInfo) {
...
+		_ = i.emitter.EmitMonikerEdge(resultSetID, monikerID)
+
+		// Link the ranges correctly to the result
+		for documentID, rangeSet := range documentReferences {
+			rangeIDs := make([]uint64, len(rangeSet))
+			index := 0
+			for rangeID := range rangeSet {
+				rangeIDs[index] = rangeID
+				index += 1
+
```
efritz, line 914:
```suggestion
			rangeIDs := make([]uint64, 0, len(rangeSet))
			for rangeID := range rangeSet {
				rangeIDs = append(rangeIDs, rangeID)
```

internal/indexer/moniker.go
tjdevries: oops, those things were to-done :laughing:

internal/indexer/indexer.go
tjdevries: If I don't typecast, then we get this problem when we try and set obj:

```
1. cannot use caseClause (variable of type ObjectLike) as types.Object value in assignment: missing method color
```

efritz: Oh it has those weird unexported methods on the interface. Gotcha.

CHANGELOG.md
```diff
@@ -11,6 +11,14 @@ All notable changes to `lsif-go` are documented in this file.
 
 ## Unreleased changes
 
+- Many issues relating to package declarations, imports and structs have been fixed.
+  - See [Package Declarations](./docs/package_declarations.md)
+  - See [Imports](./docs/imports.md)
+  - See [Structs](./docs/structs.md)
+- Additionally, package declarations are now indexed.
+
+## Fixed
```
efritz, line 20:
```suggestion
## Fixed

- Many issues relating to package declarations, imports and structs have been fixed.
  - See [Package Declarations](./docs/package_declarations.md)
  - See [Imports](./docs/imports.md)
  - See [Structs](./docs/structs.md)
- Additionally, package declarations are now indexed.
```

docs/examples/imports/.gitignore
```diff
@@ -0,0 +1 @@
+dump.lsif
```
efritz, line 1: Can just put this in root gitignore.

docs/examples/smollest/.gitignore
```diff
@@ -0,0 +1 @@
+dump.lsif
```
efritz, line 1: same

docs/imports.md
```diff
@@ -0,0 +1,48 @@
+# Imports
```
efritz, line 1: <3

efritz (approved):
The codeintel-qa pipeline passes when the indexes are generated from this branch! 🎉

PR #189, opened 2021-08-13
API docs: add tests & fix issue indexing illegal moby/moby Go code

slimsag:
https://github.com/moby/moby contains some illegal Go code (conflicting symbols defined in `_test.go` and non-`_test.go`) which confuses `go/packages` and results in some definition data not being present.

This adds test for this edge case, and makes it so we merely do not index such illegally-conflicting symbols for API docs.

Note: It is possible we will also run into this with other types of illegally conflicting symbols: this PR only handles `var` / `const` conflicts, we could run into this with type specs and function declarations theoretically. I have not handled those in this PR, is it reasonable to hope that's not in the wild, too?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ All notable changes to `lsif-go` are documented in this file.
 
 ## Unreleased changes
 
+## v1.6.6
+
+### Fixed
+
+- An issue where illegal code (conflicting test/non-test symbol names, such as in some moby/moby packages) would fail to index. [#186](https://github.com/sourcegraph/lsif-go/pull/186)
+
 ## v1.6.5
 
 ### Fixed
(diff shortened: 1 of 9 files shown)


slimsag:
Also confirmed moby does index after this:

```
$ GO111MODULE=off lsif-go
⠸ Resolving module name... WARNING: No go.mod file found in current directory.
✔ Resolving module name... Done
WARNING: No go.mod file found in current directory.
✔ Loading packages... Done
✔ Emitting documents... Done
✔ Adding import definitions... Done
✔ Indexing documentation... Done
✔ Indexing definitions... Done
✔ Indexing references... Done
✔ Linking items to definitions... Done
✔ Emitting contains relations... Done
```

efritz (approved):

PR #194, opened 2021-08-31
update sourcegraph/lib and sourcegraph/lsif-static-doc dependencies

slimsag:
Before this change, running `go test` would sometimes incorrectly delete
generated markdown and JSON files for the API docs snapshot tests. I had
thought this might be due to Go test caching, but in fact it turns out it
was just a race condition.

I had already fixed the race condition in https://github.com/sourcegraph/sourcegraph/pull/23525
but had not updated the dependency here, and as such generation of the
markdown and JSON files were racy.

@tjdevries @efritz This was the cause of the issue you were experiencing.


efritz (approved):
🚀

PR #196, opened 2021-09-01
API docs: add support for indexing code with illegal multiple main functions

slimsag:
Repositories like Go (hopefully just Go, but there are signals there may be more)
have multiple `main` functions in the same Go file, which is illegal and rejected
by the Go compiler. Additionally, it produces illegal API docs due to duplicate
path IDs:

```
t=2021-08-31T17:02:59-0700 lvl=eror msg="API docs: upload failed due to duplicate pathIDs" duplicates=1 nonDuplicates=71
t=2021-08-31T17:02:59-0700 lvl=warn msg="API docs: duplicate pathID forbidden" pathID=/illegal_multiple_mains#main
```

This adds a test case for it, and also adds support for it so the duplicative one becomes e.g. `/illegal_multiple_mains#main.<file>.<number>` similar to multiple init functions.

Helps sourcegraph/sourcegraph#24255

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to `lsif-go` are documented in this file.
 ### Fixed
 
 - An issue where API docs would not be generated for packages containing multiple `init` functions in the same file. [#195](https://github.com/sourcegraph/lsif-go/pull/195)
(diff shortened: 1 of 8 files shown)


slimsag:
@efritz gonna merge this and do a release, I think this will fix the issues [people are reporting](https://sourcegraph.slack.com/archives/C01SWEDMMGR/p1630355640012300) with e.g. these links being inconsistent / often missing API docs currently:

- https://sourcegraph.com/github.com/golang/go/-/docs/encoding/json
- https://sourcegraph.com/github.com/golang/go/-/docs/net/http

efritz (commented):
LGTM

PR #199, opened 2021-09-09
Add support for find implementations

chrismwendt:
This makes lsif-go emit interface/type implementation data necessary to support find-implementations.

- [x] List all the interfaces and concrete types for the stdlib and current project (but not other dependencies - we keep hitting an error "missing go.sum entry")
- [x] Emit a `textDocument/implementation` for each pair (interface, concrete type) that matches
- [x] Emit "interface monikers" (new term) for cross-repo find implementations
- [x] Also list other dependencies
- [x] Emit package information
- [x] Emit `textDocumentation/implementation`s for methods + cross repo

## Benchmarks

Indexing sourcegraph/sourcegraph:

- `lsif-go@master`: took 20s wall 45s user 336MB dump
- `lsif-go@implementations`: took 29s wall 69s user 339MB dump (8s to load deps, 0.5s to index implementations)

Indexing aws/aws-sdk-go:

- `lsif-go@master`: took 38s wall 106s user 2.347GB dump
- `lsif-go@implementations`: took 45s wall 123s user 2.386GB dump (1.5s to load deps, 2.7s to index implementations)

**Conclusion**: ~25% slower, ~1% bigger dumps


internal/indexer/indexer.go
```diff
@@ -854,6 +860,95 @@ func (i *Indexer) indexReferenceToExternalDefinition(p *packages.Package, docume
...
+		Logf: i.packagesLoadLogger,
+	}
+
+	deps := []*packages.Package{}
+	for _, dep := range i.dependencies {
+		fmt.Println("-----------", dep.Name)
+		name := dep.Name
+		name = strings.TrimPrefix(name, "https://")
+		name = strings.TrimPrefix(name, "https:/")
+		pkgs, err := packages.Load(config, name)
```
efritz, line 877: lol @tjdevries you were right

tjdevries: of course, but about what? :laughing: We saw this while we were testing stuff out today and figured we had to do something like this.

efritz: I gave you crap about us only calling `package.Load` once and future proofing with that cache 😃

internal/indexer/helpers_test.go
```diff
@@ -234,30 +235,9 @@ func findPackageInformationByID(w *capturingWriter, id uint64) (protocol.Package
...
-		}
-	}
-
-	return ranges
-}
-
-// findReferenceRangesByReferenceResultID returns the ranges attached to the reference result with the given
-// identifier.
-func findReferenceRangesByReferenceResultID(w *capturingWriter, id uint64) (ranges []protocol.Range) {
+func findRangesByResultID(w *capturingWriter, id uint64) (ranges []protocol.Range) {
```
tjdevries, line 240: Can any other helper functions be simplified w/ this function?

internal/indexer/indexer.go
```diff
@@ -854,6 +866,497 @@ func (i *Indexer) indexReferenceToExternalDefinition(p *packages.Package, docume
 	return rangeID, true
 }
 
+type def struct {
```
tjdevries, line 869: I had been thinking it might be nice to split up `def` into `interfaceDef` and `typeDef` (not good names) both for type safety (in terms of passing them to functions in the correct spot) but also we can probably reduce some casting if we do that.

internal/indexer/indexer.go
```diff
@@ -854,6 +866,497 @@ func (i *Indexer) indexReferenceToExternalDefinition(p *packages.Package, docume
...
+	pkg   *packages.Package
+	obj   types.Object
+	ident *ast.Ident
```
tjdevries, line 872: consider storing the method list here, so we only have to look it up once per type (we use it again later)

internal/indexer/indexer.go
```diff
@@ -854,6 +866,497 @@ func (i *Indexer) indexReferenceToExternalDefinition(p *packages.Package, docume
...
+func (i *Indexer) indexImplementations() error {
+	// Load all dependencies
+	deps, err := i.loadDependencyPackages()
+	if err != nil {
+		return err
+	}
+
+	start1 := time.Now()
+	localInterfaces, localConcreteTypes := extractTypes(i.packages)
+	remoteInterfaces, remoteConcreteTypes := extractTypes(deps)
```
tjdevries, line 925: i'd still be somewhat interested in us seeing if we can calculate the implemetations on a per-dependency basis (and then parallelize that via go funcs) and see if we get any noticeable performance improvements on large repos.

internal/indexer/indexer.go
```diff
@@ -854,6 +866,497 @@ func (i *Indexer) indexReferenceToExternalDefinition(p *packages.Package, docume
...
+		monikerKeyToID := map[string]uint64{}
+
+		emitImplementations := func(concreteTypes, interfaces []def, rightKind int, shouldInvert bool) {
+			relation := i.buildImplementationRelation(concreteTypes, interfaces)
+			leftDefs, rightDefs := concreteTypes, interfaces
+
+			if shouldInvert {
+				relation = invert(relation)
+				leftDefs, rightDefs = rightDefs, leftDefs
+			}
```
tjdevries, line 963: I have been wondering a bit about whether we should teach the backend that this language is reciprocal or whether we should actually emit both ways.

For example, if we say "struct X implements interface Y", is it always true that "interface Y is implemented by X"?

If so, then we do not need to emit everything for both directions. We can simply have:

```
X --textDocument/implement --> implementationResult --item--> Y
```

And the backend must handle the face that the reverse relationship is true as well.

cc @efritz so we can chat about this.

efritz: The relationship _should_ be symmetric (imo, just by intuition). I was chatting with @chrismwendt about this and I think it might have some tendrils in the fact that we need to emit relationships for both:

- interfaces defined in a package and a conforming struct defined in a dependency, and
- structs defined in a package and a descriptive interface of that struct defined in a dependency

internal/indexer/helpers_test.go
chrismwendt: I couldn't find a way to reuse it in 5 mins of inspection, but maybe you see one 👀

internal/indexer/indexer.go
chrismwendt: Paired on this, 2 separate types made a few call sites more type safe but bloated other parts than need to treat interfaces and types in the same way.

internal/indexer/indexer.go
chrismwendt: Are defs/refs symmetric, too? (I think so) If so, should indexers only emit one direction and the backend fill in the other? If not, there are probably some practical or principled reason(s) why that can help us decide whether or not implementations should follow the same model.

This question applies to implementation relationships within an LSIF dump, but not across dumps because cross-dump find-implementations uses monikers rather than `textDocument/implementation` et al.

tjdevries:
image

So this is one thing I'm not sure about -- putting here to talk about w/ @efritz and/or @chrismwendt at some point.

tjdevries:
This PR creates the issue: #208

Issue #213, opened 2021-11-08
Load one dependency at a time for implementations


chrismwendt:
This would reduce memory usage and obviate the `--skip-deps` flag #212 

Original motivation, OOMing in CI https://github.com/sourcegraph/sourcegraph/issues/27211

PR #216, opened 2021-11-10
Load deps in batches

chrismwendt:
Prior to this change, running lsif-go in CI (7GB RAM on GitHub Actions machines) wouldn't work because the OS OOM killed it. It needed 13+GB to work.

This change will load dependency packages in batches to avoid consuming so much memory. It removes the `--skip-deps` flag and adds a `--dep-batch-size` flag to specify the batch size.

Loading one dep at a time is ridiculously slow (would take ~22 minutes to load sourcegraph/sourcegraph's deps).

Tweaking `loadMode` reduced memory by ~10%, which is not enough.

Resolves https://github.com/sourcegraph/lsif-go/issues/213

See https://github.com/sourcegraph/sourcegraph/issues/27211 for more notes

### DIFF
--- a/cmd/lsif-go/args.go
+++ b/cmd/lsif-go/args.go
@@ -26,7 +26,7 @@ var (
 	verbosity        int
 	noOutput         bool
 	noAnimation      bool
-	skipDeps         bool
+	depBatchSize     int
 )
 
 func init() {
@@ -45,13 +45,12 @@ func init() {
 	app.Flag("repository-remote", "Specifies the canonical name of the repository remote.").Default(defaultRepositoryRemote.Value()).StringVar(&repositoryRemote)
 	app.Flag("module-version", "Specifies the version of the module defined by module-root.").Default(defaultModuleVersion.Value()).StringVar(&moduleVersion)
 
-	// Feature options
-	app.Flag("skip-deps", "Do not load depedencies - reduces memory usage but omits interface implementation data from deps.").Default("true").BoolVar(&skipDeps)
-
 	// Verbosity options
 	app.Flag("quiet", "Do not output to stdout or stderr.").Short('q').Default("false").BoolVar(&noOutput)
 	app.Flag("verbose", "Output debug logs.").Short('v').CounterVar(&verbosity)
 	app.Flag("no-animation", "Do not animate output.").Default("false").BoolVar(&noAnimation)
+
+	app.Flag("dep-batch-size", "How many dependencies to load at once to limit memory usage (e.g. 100).").Default("0").IntVar(&depBatchSize)
 }
 
 func parseArgs(args []string) (err error) {
--- a/cmd/lsif-go/args.go
+++ b/cmd/lsif-go/args.go
@@ -50,7 +50,7 @@ func init() {
 	app.Flag("verbose", "Output debug logs.").Short('v').CounterVar(&verbosity)
 	app.Flag("no-animation", "Do not animate output.").Default("false").BoolVar(&noAnimation)
 
-	app.Flag("dep-batch-size", "How many dependencies to load at once to limit memory usage (e.g. 100).").Default("0").IntVar(&depBatchSize)
+	app.Flag("dep-batch-size", "How many dependencies to load at once to limit memory usage (e.g. 100). 0 means load all at once.").Default("0").IntVar(&depBatchSize)
 }
 
 func parseArgs(args []string) (err error) {
--- a/cmd/lsif-go/index.go
+++ b/cmd/lsif-go/index.go
@@ -12,7 +12,7 @@ import (
 	"github.com/sourcegraph/sourcegraph/lib/codeintel/lsif/protocol/writer"
 )
 
-func writeIndex(repositoryRoot, repositoryRemote, projectRoot, moduleName, moduleVersion string, dependencies map[string]gomod.GoModule, projectDependencies []string, outFile string, outputOptions output.Options, skipDeps bool) error {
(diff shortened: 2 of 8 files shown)


efritz:
> Loading one dep at a time is ridiculously slow (would take ~22 minutes to load sourcegraph/sourcegraph's deps).

Do we have any insights on how to choose batch sizes?

cmd/lsif-go/args.go
```diff
@@ -45,13 +45,12 @@ func init() {
...
 
-	// Feature options
-	app.Flag("skip-deps", "Do not load depedencies - reduces memory usage but omits interface implementation data from deps.").Default("true").BoolVar(&skipDeps)
-
 	// Verbosity options
 	app.Flag("quiet", "Do not output to stdout or stderr.").Short('q').Default("false").BoolVar(&noOutput)
 	app.Flag("verbose", "Output debug logs.").Short('v').CounterVar(&verbosity)
 	app.Flag("no-animation", "Do not animate output.").Default("false").BoolVar(&noAnimation)
+
+	app.Flag("dep-batch-size", "How many dependencies to load at once to limit memory usage (e.g. 100).").Default("0").IntVar(&depBatchSize)
```
efritz, line 53: If zero means unbounded could we document that?

internal/indexer/implementation.go
```diff
@@ -348,7 +376,37 @@ func (i *Indexer) extractInterfacesAndConcreteTypes(pkgs []*packages.Package) (i
...
+	for ix, pkgName := range pkgNames {
+		pkgBatch = append(pkgBatch, pkgName)
+
+		if i.depBatchSize != 0 && ix%i.depBatchSize == 0 {
+			err := batch(pkgBatch)
+			runtime.GC() // Prevent a garbage pile
+			if err != nil {
+				return nil, nil, err
+			}
+			pkgBatch = []string{}
```
efritz, line 401:
```suggestion
			pkgBatch = pkgBatch[:0]
```

internal/indexer/implementation.go
```diff
@@ -348,7 +376,37 @@ func (i *Indexer) extractInterfacesAndConcreteTypes(pkgs []*packages.Package) (i
...
+		}
+	}
+	err = batch(pkgBatch)
+	if err != nil {
```
efritz, line 405:
```suggestion
	if err := batch(pkgBatch); err != nil {
```

internal/indexer/implementation.go
efritz: Yeah it just resets the slice to the beginning of the array (all capacity, no length). Reuses the same memory over and over while it's live.

efritz: Note that this trick only works if you're not using that array concurrently with something else that expects stable objects - this is fine since we're batching synchronously in a loop and not tasking things out to goroutines.

cmd/lsif-go/index.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/sourcegraph/sourcegraph/lib/codeintel/lsif/protocol/writer"
 )
 
-func writeIndex(repositoryRoot, repositoryRemote, projectRoot, moduleName, moduleVersion string, dependencies map[string]gomod.GoModule, projectDependencies []string, outFile string, outputOptions output.Options, skipDeps bool) error {
+func writeIndex(repositoryRoot, repositoryRemote, projectRoot, moduleName, moduleVersion string, dependencies map[string]gomod.GoModule, projectDependencies []string, outFile string, outputOptions output.Options, depBatchSize int) error {
```
varungandhi-src, line 15: Would it make sense to break these up into structs? Some bits here are information about the project, and some bits are related to I/O.

chrismwendt:
I documented the recommended batch size in the README:

> If lsif-go is using too much memory, try setting `--dep-batch-size=100` to only load 100 dependencies into memory at once (~1GB overhead). Lowering the batch size will decrease the overhead further, but increase the runtime a lot more because loading a batch has a fixed cost of ~500ms and each additional package loaded within a batch only adds ~10ms.

cmd/lsif-go/index.go
chrismwendt: That could help categorize them - cc @tjdevries who's considering doing some refactoring.

internal/indexer/implementation.go
```diff
@@ -348,7 +376,36 @@ func (i *Indexer) extractInterfacesAndConcreteTypes(pkgs []*packages.Package) (i
 		}
 	}
 
-	return interfaces, concreteTypes
+	batch := func(pkgBatch []string) error {
+		pkgs, err := i.loadPackage(true, pkgBatch...)
```
varungandhi-src, line 380: Is `loadPackage` ever called with `deduplicate = false` in non-test code? It looks like the only place that happens is when `loadPackages` is called with `deduplicate = false` [in a test](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/lsif-go%24+loadPackages%28&patternType=literal&case=yes). Would it make sense to remove the argument and test only the `deduplicate = true` case?

chrismwendt: Hmm, this does seem unused. Looks like this originated in https://github.com/sourcegraph/lsif-go/pull/138/files?w=1#diff-87c2475baa87005bd3a3001ec33f201feea23caddbbf5a2f249238a2bd32026dR122 with the comment "deduplicate should be true in all cases except TestIndexer_shouldVisitPackage." @slimsag might be able to shed light on this one 💡

slimsag: I don't recall, I could dig in if needed but I'd say go with your gut/instinct here!

PR #218, opened 2021-11-13
Bump checkout action to v2.

varungandhi-src:
Previously, we discovered a problem with the sg/sg
repo, only when using v2 of the checkout action.
However, this problem was not caught earlier because
lsif-go was using v1. We should keep the action versions
in sync across repos.

### DIFF
--- a/.github/workflows/lsif.yml
+++ b/.github/workflows/lsif.yml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     container: sourcegraph/lsif-go
     steps:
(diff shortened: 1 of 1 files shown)


varungandhi-src:
(Will merge this on Monday.)

tjdevries:
@varungandhi-src do you need anything else to merge or is this fine?

varungandhi-src:
Oh, I don't think I need anything else. I just didn't want to merge on the weekend in case there was any fallout, since it wasn't pressing.

PR #219, opened 2021-11-13
DO NOT MERGE Bump golang test

varungandhi-src:


varungandhi-src:
Hmm, weird, there is no failure here, unlike in the sourcegraph/sourcegraph CI earlier: https://github.com/sourcegraph/sourcegraph/runs/4169610043?check_suite_focus=true

PR #220, opened 2021-11-16
feat: greatly reduce memory usage by random guessing

tjdevries:
can remove prints or add them to stats

### DIFF
--- a/internal/indexer/indexer.go
+++ b/internal/indexer/indexer.go
@@ -249,21 +249,19 @@ func (i *Indexer) loadPackages(deduplicate bool) error {
 }
 
 func (i *Indexer) loadPackage(deduplicate bool, patterns ...string) ([]*packages.Package, error) {
-	loadProject := len(patterns) == 1 && patterns[0] == "./..."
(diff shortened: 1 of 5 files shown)


internal/indexer/indexer.go
```diff
@@ -248,29 +249,45 @@ func (i *Indexer) loadPackages(deduplicate bool) error {
...
-	if len(patterns) == 1 && patterns[0] == "./..." && i.packages != nil {
+	loadProject := len(patterns) == 1 && patterns[0] == "./..."
+	if loadProject && i.packages != nil {
 		return i.packages, nil
 	}
 
+	loadTests := false
+	if loadProject {
+		loadTests = true
+	}
```
varungandhi-src, line 260:
```suggestion
	loadTests := loadProject
```

tjdevries: changed var name, but did this. ty

tjdevries:
I had some mixed stuff here between updating benchmarks and test loading. Moved to #221 for reducing memory usage

PR #221, opened 2021-11-16
feat: reduce memory usage by not loading tests of dependencies

tjdevries:


chrismwendt (approved):
Good idea to try skipping tests 😺 Stats in the description would be nice for future reference

internal/indexer/indexer.go
```diff
@@ -248,15 +249,23 @@ func (i *Indexer) loadPackages(deduplicate bool) error {
...
+		Dir:  i.projectRoot,
+		Logf: i.packagesLoadLogger,
+
+		// Only load tests for the current project.
+		// This greatly reduces memory usage when loading dependencies
+		Tests: isLoadingProject,
+
+		// CGO_ENABLED=0 makes sure that we can handle files with assembly
+		// and other problems that may occur (unsure of exact reasons at time of writing)
+		Env: append(os.Environ(), "CGO_ENABLED=0"),
```
chrismwendt, line 268: This looks unrelated to the intent of the PR and might be causing https://github.com/sourcegraph/lsif-go/issues/227

It also prevents the user from configuring `CGO_ENABLED` on the command line.

@tjdevries should we revert this piece?

PR #222, opened 2021-11-24
ahaha this actually passes test

tjdevries:


chrismwendt:
I'm a bit surprised this works with so few changes to the main indexing code 😃 Neat

How does this compare to `--dep-batch-size=100` in terms of speed and memory?

Does this make it more or less difficult to add features like call graph and type definition?

Do we have examples where this reduction in memory usage is necessary? Maybe on k8s in a GitHub Action with only 12GB of memory? [`env GOGC=20`](https://dave.cheney.net/2015/11/29/a-whirlwind-tour-of-gos-runtime-environment-variables) (anything lower than the default of 100) would also help a lot.

tjdevries:
Will be solved in https://github.com/sourcegraph/scip-go

Issue #227, opened 2022-02-16
Indexing CGo?


colinking:
We're running into an issue with `v1.7.5` because of CGo:

```
error: failed to index: failed to load packages: packages.Load: err: exit status 1: stderr: go build github.com/pganalyze/pg_query_go/v2/parser: build constraints exclude all Go files in /go/pkg/mod/github.com/pganalyze/pg_query_go/v2@v2.1.0/parser
```

Looks related to https://github.com/sourcegraph/lsif-go/pull/221/files which sets `CGO_ENABLED=0` when loading packages. For now, we can lock to `1.7.4`.

PR #233, opened 2022-02-18
Do not disable cgo (rm CGO_ENABLED=0)

chrismwendt:
You can disable it with `env CGO_ENABLED=0 lsif-go` if necessary.

Fixes https://github.com/sourcegraph/sourcegraph/issues/33285

Might fix https://github.com/sourcegraph/sourcegraph/pull/30307#issuecomment-1041917665. I was unable to repro the error, but I was able to determine that `binding.go` is now included in the output whereas before this change it wasn't.

Might fix https://github.com/sourcegraph/lsif-go/issues/227 for the same reason.

I didn't notice any negative consequences of this change:

- gorillla/mux still works
- sourcegraph/src-cli still works
- sourcegraph/sourcegraph/lib still works without the [workaround](https://github.com/sourcegraph/sourcegraph/blob/672ea0c167b971fc01cc98ba55a11451401f77dd/lib/codeintel/repro_lang/src/workaround.go). Olaf had an [issue](https://github.com/sourcegraph/sourcegraph/pull/30307#issuecomment-1041917665) but I was unable to repro.
- google/gvisor (with C files) still doesn't work due to having `.c` files in Go packages without `import "C"`
- gonum/gonum (with assembly files) still works, same set of files ended up in the LSIF dump
- A fresh repo with https://riptutorial.com/go/example/21315/cgo--first-steps-tutorial emits more info than before this change

### DIFF
--- a/internal/indexer/indexer.go
+++ b/internal/indexer/indexer.go
@@ -275,10 +275,6 @@ func (i *Indexer) loadPackage(deduplicate bool, patterns ...string) ([]*packages
 		// Only load tests for the current project.
 		// This greatly reduces memory usage when loading dependencies
 		Tests: isLoadingProject,
-
-		// CGO_ENABLED=0 makes sure that we can handle files with assembly
-		// and other problems that may occur (unsure of exact reasons at time of writing)
-		// Env: append(os.Environ(), "CGO_ENABLED=0"),
 	}
 
 	// Make sure we only load packages once per execution.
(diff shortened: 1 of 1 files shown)


varungandhi-src:
Could you elaborate on how this fixes https://github.com/sourcegraph/sourcegraph/issues/33285?

varungandhi-src (approved):
Nvm, looked at the other issue as well. Makes sense.

PR #237, opened 2022-03-03
Add changelog entry for v1.7.6.

varungandhi-src:
@chrismwendt: is there a nice description for when the panics could happen that I could add to the changelog here? I couldn't quite figure out what user code would trigger the issues fixed in the patch.

### Test plan

Not needed since we're only adding a changelog entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@
 
 All notable changes to `lsif-go` are documented in this file.
 
+## v1.7.6
+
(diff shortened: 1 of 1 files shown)


chrismwendt:
There isn't much else to say except that removing `CGO_ENABLED=0` uncovered a few nil accesses that were causing SEGFAULTs.

chrismwendt (approved):

PR #245, opened 2022-05-10
Support `darwin/arm64` target

michaellzc:
- Add apple silicon support
- Fix buildkite pipeline
- Upgrade goreleaser to the latest 1.x release (there's no darwin_arm built for pre 1.x goreleaser)

### Test plan

🤞 

it passed
```sh
goreleaser release --snapshot --rm-dist
```


efritz (approved):

.tool-versions
```diff
@@ -0,0 +1 @@
+golang 1.18.1
```
Strum355, line 1: Is this right, given we seem to be using 1.16 in release.yml? 
which also makes me wonder, is 1.16 right? :P @tjdevries @efritz @chrismwendt

michaellzc: should we just bump `go-version` in release.yml to 1.18 as well?

Strum355: That version has implications on lsif-go itself afaik

michaellzc: @Strum355 reverted back to 1.16 for buildkite pipeline and it also supports m1 chip ;)

Strum355 (approved):
:nice:

varungandhi-src:
Meta question: Why do we have both buildkite and GitHub Actions in this repo? Could we do with just one instead of both?

PR #248, opened 2022-05-11
Upgrade 1.18

olafurpg:
This PR upgrades lsif-go to build with Go v1.18, which added support for generics. A quick experiment reveals that navigation should work correctly in files that are using generics excluding type parameters themselves because we don't render correct monikers for those yet.

### Test plan

See the CI go green.


internal/gomod/stdlib.go
```diff
@@ -1,21 +1,11 @@
...
 // or not. At this point, it checks whether the package name is one of those
 // that is found from running "go list std" in the latest released go version.
 func isStandardlibPackge(pkg string) bool {
-	// For certain items in the Universe scope, they have no package.
-	// Because of this, we put them in the "builtin" package, because
-	// that's the name of the file in the go standard library that defines
-	// these items.
-	//
-	// An example would be `error` or `error.Error()`
-	if pkg == "builtin" {
```
Strum355, line 15: Is this meant to be removed?

olafurpg: This file is auto-generated by `./scripts/gen_stdlib_map.sh`.

Strum355 (approved):
Hopefully this improves handling of files with generics 🤞

PR #256, opened 2022-09-01
Disable animation by default

olafurpg:
Animation currently breaks manual indexing of sourcegraph/sourcegraph
because it blocks a prompt from git about trusting a fingerprint.
Until we figure out a way to enable stdin together with animation, it's
best to disable animation so that customers don't experienc buggy
behavior out of the box.

### Test plan

I built the binary locally and manually verified that I was able to index sourcegraph/sourcegraph with the command `lsif-go` (no flags). Before this commit, I needed to manually add `--no-animation`, which is not obvious if you're a customer and trying to index a codebase.

### DIFF
--- a/cmd/lsif-go/args.go
+++ b/cmd/lsif-go/args.go
@@ -25,7 +25,7 @@ var (
 	moduleVersion         string
 	verbosity             int
 	noOutput              bool
-	noAnimation           bool
+	animation             bool
 	depBatchSize          int
 	enableApiDocs         bool
 	enableImplementations bool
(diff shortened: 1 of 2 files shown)


cmd/lsif-go/args.go
```diff
@@ -50,7 +50,7 @@ func init() {
 	// Verbosity options
 	app.Flag("quiet", "Do not output to stdout or stderr.").Short('q').Default("false").BoolVar(&noOutput)
 	app.Flag("verbose", "Output debug logs.").Short('v').CounterVar(&verbosity)
-	app.Flag("no-animation", "Do not animate output.").Default("false").BoolVar(&noAnimation)
+	app.Flag("no-animation", "Do not animate output.").Default("true").BoolVar(&noAnimation)
```
olafurpg, line 53: I wasn't able to figure out how to set `no-animation` to true if you want to force animation. I don't consider it a blocking issue for this PR, but it would be nice to allow users to opt-into animation (although I doubt anyone will bother).

mrnugget: You won't believe this (I had to do a double take) but apparently the library used here, `kingping`, [creates a `--no-<flag>` complement for each boolean flag](https://github.com/alecthomas/kingpin#boolean-values), so... yes:

```
./lsif-go --no-no-animation
```

mrnugget (approved):

cmd/lsif-go/args.go
olafurpg: Very nice. This is quite common among cli parser libraries, but I didn't give it a try. Just updated the flag to allow `--animation`

olafurpg:
Closing this PR because it's from my fork. Will open a new PR from the main repo

PR #259, opened 2022-11-02
init scip: for visibility

tjdevries:
Just pushing so @olafurpg can look at the code

### Test plan

- [ ] I will write tests and what not


internal/index/scip.go
```diff
@@ -0,0 +1,432 @@
...
+				Range:       scipRange(position, ref),
+				Symbol:      scipSymbol(findModuleForObj(f.deps, ref), ref),
+				SymbolRoles: int32(scip.SymbolRole_ReadAccess),
+			})
+		}
+
+		fmt.Println("Definition: ", def)
+		fmt.Println("References: ", ref)
+	default:
+		fmt.Printf("unhandled: %T %v\n", n, n)
```
olafurpg, line 392: Is there a way to fallback to the default visitor that continues recursing on the AST?

tjdevries: well it will continue to recurse the AST even with this strategy -- I just am letting myself know that we didn't handle something

tjdevries:
moved to https://github.com/sourcegraph/scip-go

PR #260, opened 2022-11-07
gomod: special case listing the stdlib

keegancsmith:
Before this commit lsif-go fails on the standard library for go. When running the "go list all" command fails to run with an error message like:

```
cannot find package "." in:
  path/to/a/vendor/dir
...
```

This might be an issue in upstream go list. However, you normally do need special handling of the go stdlib tool. In our case we can just skip listing dependencies since it is the standard library.

Test Plan: We eyeballed the lsif dump after running on stdlib, and it looked legitimate. However, we don't know much about lsif.

Co-authored-by: @stefanhengl

### DIFF
--- a/cmd/lsif-go/main.go
+++ b/cmd/lsif-go/main.go
@@ -51,7 +51,7 @@ func mainErr() (err error) {
 		ShowAnimations: animation,
 	}
 
-	moduleName, err := gomod.ModuleName(moduleRoot, repositoryRemote, outputOptions)
+	moduleName, isStdLib, err := gomod.ModuleName(moduleRoot, repositoryRemote, outputOptions)
 	if err != nil {
 		return fmt.Errorf("failed to infer module name: %v", err)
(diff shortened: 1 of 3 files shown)


efritz (approved):
Nvm, this could unblock indexing the gostd lib to give page ranks before GHU.

tjdevries:
Also, I'm trying to re-use as much of the original stuff we have (detecting stdlib, package, etc.), so improvements to this part of the code are useful. I'll take a look right now

tjdevries (approved):
Changes seem fine -- I'm not sure exactly how you were running it, but if it's working for you I think we can merge.

keegancsmith:
> Changes seem fine -- I'm not sure exactly how you were running it, but if it's working for you I think we can merge.

Just running lsif-go in the golang/go src dir. If you see our auto indexing jobs on sg.com they are failing for the src dir.

PR #267, opened 2023-09-14
Upgrade Go to 1.21

olleman42:
This is to increase compatability with Go projects that use new std lib features such as `slices`

### Test plan

The CI should pass.


raybarrera:
Structured logging also causes issues without this PR.

varungandhi-src:
Apologies for not archiving this repo earlier. Please use [scip-go](https://github.com/sourcegraph/scip-go) instead of lsif-go; lsif-go is no longer supported.

